PR #25, opened 2017-11-10
adding fallback option when druid returns empty lookup

pavanab4u:


core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -241,16 +255,16 @@ case class MultiEngineQuery(drivingQuery: Query,
     val drivingResult = executorsMap(drivingQuery.engine).execute(drivingQuery, rowList.asInstanceOf[IndexedRowList], queryAttributes)
     val drivingQueryEndTime = System.currentTimeMillis()
     engineQueryStats.addStat(EngineQueryStat(drivingQuery.engine, drivingQueryStartTime, drivingQueryEndTime))
-    val result = if (drivingResult._1.keys.isEmpty) drivingResult else subsequentQueries.foldLeft(drivingResult) {
+    val result = if (drivingResult.rowList.asInstanceOf[IndexedRowList].keys.isEmpty) drivingResult else subsequentQueries.foldLeft(drivingResult) {
```
patelh: is this asInstanceOf safe?  why do u need that to get to keys.isEmpty?

core/src/main/scala/com/yahoo/maha/core/query/QueryExecutor.scala
```diff
@@ -16,7 +16,7 @@ import scala.concurrent.Future
 trait QueryExecutor {
   def engine: Engine
   def acceptEngine(engine: Engine) : Boolean = engine == this.engine
-  def execute[T <: RowList](query: Query, rowList: T, queryAttributes: QueryAttributes) : (T, QueryAttributes)
+  def execute[T <: RowList](query: Query, rowList: T, queryAttributes: QueryAttributes) : QueryResult
```
patelh: Maybe make the QueryResult typed with T, QueryResult[T]

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -837,8 +851,19 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
           } else {
             if (factBestCandidateOption.isDefined) {
               val query = getDimFactQuery(bestDimCandidates, factBestCandidateOption.get, requestModel, queryAttributes)
+              val fallbackQueryOption: Option[(Query, RowList)] = {
+                if (requestModel.bestCandidates.isDefined
+                  && requestModel.bestCandidates.get.facts.values.map(_.fact.engine).toSet.size > 1
+                  && requestModel.forceQueryEngine.isEmpty) {
+                  val (newFactBestCandidateOption, newBestDimCandidates) = findBestCandidates(requestModel, Set(factBestCandidateOption.get.fact.engine))
```
patelh: If there is no other fallback engine, this will fail, you probably don't want to have a fallback in that case.

pavanab4u: that condition exists 
`requestModel.bestCandidates.get.facts.values.map(_.fact.engine).toSet.size > 1`

PR #28, opened 2017-11-15
toDate no longer defaults to the future

ryankwagner:


patelh:
This won't fix missing segments from today.

druid/src/main/scala/com/yahoo/maha/executor/druid/DruidQueryExecutor.scala
```diff
@@ -269,7 +269,7 @@ class DruidQueryExecutor(config:DruidQueryExecutorConfig , lifecycleListener: Ex
 
   def checkUncoveredIntervals(query : Query, response : Response) : Unit = {
     val requestModel = query.queryContext.requestModel
-    val latestDate : DateTime = requestModel.queryGrain.getOrElse(DailyGrain).fromFormattedString(requestModel.getMostRecentRequestedDate())
+    val latestDate : DateTime = FilterDruid.getMaxDate(requestModel.utcTimeDayFilter, DailyGrain)
```
patelh: After now may be too restrictive.  It should be before today, now is before current system milliseconds.

PR #30, opened 2017-11-16
Engine specific filter validation

pavanab4u:


pavanab4u:
@patelh please review

core/src/main/scala/com/yahoo/maha/core/Engine.scala
```diff
@@ -45,3 +48,26 @@ trait WithOracleEngine extends EngineRequirement {
 trait WithDruidEngine extends EngineRequirement {
   final val engine: Engine = DruidEngine
 }
+
+object EngineConstraint {
```
patelh: I thought we talked about putting this on the query generator?

patelh:
lgtm

PR #31, opened 2017-11-18
Add config in service for interval fall back feature

pranavbhole:


druid/src/main/scala/com/yahoo/maha/executor/druid/DruidQueryExecutor.scala
```diff
@@ -267,10 +264,13 @@ class DruidQueryExecutor(config:DruidQueryExecutorConfig , lifecycleListener: Ex
...
 
-  def checkUncoveredIntervals(query : Query, response : Response) : Unit = {
+  def checkUncoveredIntervals(query : Query, response : Response, config: DruidQueryExecutorConfig) : Unit = {
     val requestModel = query.queryContext.requestModel
     val latestDate : DateTime = FilterDruid.getMaxDate(requestModel.utcTimeDayFilter, DailyGrain)
-    if(latestDate.isBeforeNow() && response.getHeaders().containsKey(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT) && response.getHeader(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT).contains(DruidQueryExecutor.UNCOVERED_INTERVAL_VALUE)){
+    if (latestDate.isBeforeNow()
+      && response.getHeaders().containsKey(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT)
+      && response.getHeader(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT).contains(DruidQueryExecutor.UNCOVERED_INTERVAL_VALUE)
+      && config.enableFallbackOnUncoveredIntervals) {
```
pavanab4u: add this as first condition so if this is false then it can skip other 3 checks

PR #37, opened 2017-12-01
Core coverage

ryankwagner:


core/src/main/scala/com/yahoo/maha/jdbc/package.scala
```diff
@@ -6,94 +6,19 @@
...
 
-import java.sql.{Date, ResultSet, Timestamp}
-import java.text.SimpleDateFormat
+import java.sql.ResultSet
 
 import scala.util.Try
 
 package object jdbc {
-
-  type Seq[+A] = scala.collection.immutable.Seq[A]
```
pranavbhole: isnt it used  anywhere?

ryankwagner: It wasn't used in tests, or even the maha-api-example file that was attempting to use it.  I built it, then I rebuilt CDW with this change, and nothing broke.  A lot of this was extra stuff from the GitHub credited in the jdbcConnection file.

ryankwagner: Did manage to find an issue with it.  The question is if we want to keep that specific definition inside that file or move it elsewhere.  Does it belong there?

PR #43, opened 2017-12-05
Add GetTotalRowsRequest to maha core utils

ryankwagner:
Currently, the function is detached from the pipeline, and is a simple standalone function.
Additionally, added tests to DaysUtils for scoverage (67% to ~90%).


core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,96 @@
...
+import com.yahoo.maha.core.query.oracle.OracleQueryGenerator
+import com.yahoo.maha.core.registry.Registry
+import com.yahoo.maha.core.request.{Field, ReportingRequest}
+
+import scala.util.Try
+
+object GetTotalRowsRequest {
+  def updateRowList(rowList: RowList) : Unit = {
+    val row = rowList.newRow
+    println(rowList.columnNames)
```
patelh: You need this?

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  def getTotalRows(request: ReportingRequest, sourcePipeline: QueryPipeline, registry: Registry) : Try[Int] = {
+    Try {
+      val totalRowsRequest: ReportingRequest = getTotalRowsRequest(request, sourcePipeline).get
+      val model: RequestModel = RequestModel.from(totalRowsRequest, registry).get
+      val maxRows: Int = DruidQueryGenerator.defaultMaximumMaxRows
+      assert(model.maxRows <= maxRows, throw new Exception(s"Value of ${model.maxRows} exceeds posted limit of $maxRows"))
+      val queryContext: QueryExecutorContext = getQueryExecutorContext
```
patelh: You don't want to create a new context every time.  Why not pass it in?

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,96 @@
...
+
+      val rowListAttempt = sourcePipeline.execute(queryContext)
```
patelh: This looks wrong, why would you want to execute the source pipeline?  Why aren't we creating a new pipeline from the new request model created from the total rows reporting request?

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,96 @@
...
+      assert(rowListAttempt.isSuccess, "Failed to get valid executor and row list")
+
+      //Can fail back in getValue exception.
+      var result = 0
+      rowListAttempt.get._1.foreach(input => {
+        result += input.aliasMap(OracleQueryGenerator.ROW_COUNT_ALIAS).toString.toInt
```
patelh: Probably need to check the number of rows returned.  Add more debug logging if debug is enabled.

patelh: Shouldn't we determine which engine the request was executed and based on that parse the result?  Or are we expecting all the engines to just use the same alias/name?

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
ryankwagner: Right now I am testing on the generator side, but I will switch my testing over to executor side so that the context is available to pass in.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
ryankwagner: Believe only Oracle has this alias, and all others have nothing at all.  I can add in a check to make sure it is running an Oracle query.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+        , forceFactDriven = false
+        , paginationStartIndex = 0
+        , rowsPerPage = request.rowsPerPage
+      )
+    }
+  }
+
+  def getTotalRows(request: RequestModel, sourcePipeline: QueryPipeline, registry: Registry, queryContext: QueryExecutorContext)(implicit queryGeneratorRegistry: QueryGeneratorRegistry) : Try[Int] = {
+    Try {
+      val totalRowsRequest: ReportingRequest = getTotalRowsRequest(request.reportingRequest, sourcePipeline).get
```
patelh: This is an unsafe get.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+      val model: RequestModel = RequestModel.from(totalRowsRequest, registry).get
```
patelh: Another unsafe get.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+      val maxRows: Int = DruidQueryGenerator.defaultMaximumMaxRows
+      assert(model.maxRows <= maxRows, throw new Exception(s"Value of ${model.maxRows} exceeds posted limit of $maxRows"))
```
patelh: Why not just set the max rows to default value since the result really should just be a single row to get the total rows count.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+
+      val queryPipelineFactory = new DefaultQueryPipelineFactory()
+
+      val requestPipelineTry = queryPipelineFactory.from(model, QueryAttributes.empty)
+      val rowListAttempt = requestPipelineTry.toOption.get.execute(queryContext)
+      assert(rowListAttempt.isSuccess, "Failed to get valid executor and row list")
```
patelh: Maybe use require instead of assert since that what we do everywhere else.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+    Try {
+      val totalRowsRequest: ReportingRequest = getTotalRowsRequest(request.reportingRequest, sourcePipeline).get
+      val model: RequestModel = RequestModel.from(totalRowsRequest, registry).get
+      val maxRows: Int = DruidQueryGenerator.defaultMaximumMaxRows
+      assert(model.maxRows <= maxRows, throw new Exception(s"Value of ${model.maxRows} exceeds posted limit of $maxRows"))
+
+      val queryPipelineFactory = new DefaultQueryPipelineFactory()
+
+      val requestPipelineTry = queryPipelineFactory.from(model, QueryAttributes.empty)
+      val rowListAttempt = requestPipelineTry.toOption.get.execute(queryContext)
```
patelh: Another unsafe get.

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+      assert(rowListAttempt.isSuccess, "Failed to get valid executor and row list")
+
+      //Can fail back in getValue exception.
+      var result = 0
+      rowListAttempt.get._1.foreach(input => {
+        assert(input.aliasMap.contains(OracleQueryGenerator.ROW_COUNT_ALIAS), "TOTALROWS not defined in alias map, only valid in Oracle Queries")
```
patelh: How do we really want to handle this kind of error?

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+        , rowsPerPage = request.rowsPerPage
+      )
+    }
+  }
+
+  def getTotalRows(request: RequestModel, sourcePipeline: QueryPipeline, registry: Registry, queryContext: QueryExecutorContext)(implicit queryGeneratorRegistry: QueryGeneratorRegistry) : Try[Int] = {
+    Try {
+      val totalRowsRequest: ReportingRequest = getTotalRowsRequest(request.reportingRequest, sourcePipeline).get
+      val model: RequestModel = RequestModel.from(totalRowsRequest, registry).get
+      val maxRows: Int = DruidQueryGenerator.defaultMaximumMaxRows
```
patelh: Why are we using the value from DruidQueryGenerator here?  Do we even need this?  Aren't we expecting to just get 1 row returned with the # of total rows?

core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -0,0 +1,64 @@
...
+        , forceDimensionDriven = true
+        , forceFactDriven = false
+        , paginationStartIndex = 0
+        , rowsPerPage = request.rowsPerPage
+      )
+    }
+  }
+
+  def getTotalRows(request: RequestModel, sourcePipeline: QueryPipeline, registry: Registry, queryContext: QueryExecutorContext)(implicit queryGeneratorRegistry: QueryGeneratorRegistry) : Try[Int] = {
+    Try {
```
patelh: Consider changing this to a for {} yield {} since you have a bunch of try's within this try.

pranavbhole:
👍

PR #51, opened 2018-01-02
totalrow request and test fixes

ryankwagner:


core/src/main/scala/com/yahoo/maha/utils/GetTotalRowsRequest.scala
```diff
@@ -54,12 +54,13 @@ object GetTotalRowsRequest extends Logging {
       var result = 0
       rowListAttempt.get._1.foreach(input => {
         require(input.aliasMap.contains(OracleQueryGenerator.ROW_COUNT_ALIAS), "TOTALROWS not defined in alias map, only valid in Oracle Queries")
-        val current_totalrows = input.aliasMap(OracleQueryGenerator.ROW_COUNT_ALIAS).toString.toInt
+        val totalrow_col_num = input.aliasMap(OracleQueryGenerator.ROW_COUNT_ALIAS)
```
pranavbhole: Can you print the exception in line 41 and 49

pranavbhole:
👍

PR #56, opened 2018-01-05
Adding PrestoQueryExecutor

panditsurabhi:


presto/src/test/scala/com/yahoo/maha/executor/PrestoQueryExecutorTest.scala
```diff
@@ -0,0 +1,577 @@
...
+  private var dataSource: Option[HikariDataSource] = None
+  private var jdbcConnection: Option[JdbcConnection] = None
+  private var prestoQueryExecutor : Option[PrestoQueryExecutor] = None
+  private val queryExecutorContext : QueryExecutorContext = new QueryExecutorContext
+  private val staticTimestamp = new Timestamp(System.currentTimeMillis())
+  private val staticTimestamp2 = new Timestamp(System.currentTimeMillis() + 1)
+
+  override protected def beforeAll(): Unit = {
+    val config = new HikariConfig()
+    config.setJdbcUrl("jdbc:h2:mem:" + UUID.randomUUID().toString.replace("-",
```
pranavbhole: is there any Presto Mode in h2? how can we test presto syntax here?

panditsurabhi: Presto allows ANSI SQL, which is what default H2 mode is. There is no separate mode for Presto.

PR #68, opened 2018-01-17
Fix join condition for different data types

panditsurabhi:


core/src/main/scala/com/yahoo/maha/core/query/presto/PrestoQueryGenerator.scala
```diff
@@ -491,15 +491,24 @@ class PrestoQueryGenerator(partitionColumnRenderer:PartitionColumnRenderer, udfS
       } else {
         "LEFT OUTER JOIN"
       }
-      // pkColName ?? alias ?? cc3_id
+
+      // Columns on which join is done must be of the same type. Add explicit CAST otherwise.
+      val joinCondition = {
+        if (fkCol.dataType.getClass.equals(dimBundle.dim.columnsByNameMap(pkColName).dataType.getClass)) {
```
pranavbhole: dimBundle.dim.columnsByNameMap(pkColName).dataType is unsafe, put require and make sure map has value

core/src/main/scala/com/yahoo/maha/core/query/presto/PrestoQueryGenerator.scala
```diff
@@ -491,15 +491,24 @@ class PrestoQueryGenerator(partitionColumnRenderer:PartitionColumnRenderer, udfS
...
+          s"$factViewAlias.$renderedFactFk = $dimAlias.${dimAlias}_$pkColName"
+        } else {
+          s"CAST($factViewAlias.$renderedFactFk AS VARCHAR) = CAST($dimAlias.${dimAlias}_$pkColName AS VARCHAR)"
```
pranavbhole: I am not sure how will it affect performance as it will cast all the keys in right and left table

core/src/test/scala/com/yahoo/maha/core/query/presto/BasePrestoQueryGeneratorTest.scala
```diff
@@ -45,7 +45,7 @@ trait BasePrestoQueryGeneratorTest
       Fact.newFact(
         "s_stats_fact", DailyGrain, PrestoEngine, Set(AdvertiserSchema),
         Set(
-          DimCol("campaign_id", IntType(), annotations = Set(ForeignKey("campaign")))
+          DimCol("campaign_id", StrType(), annotations = Set(ForeignKey("campaign")))
```
pranavbhole: why is ID str ?

core/src/main/scala/com/yahoo/maha/core/query/presto/PrestoQueryGenerator.scala
panditsurabhi: Added require

core/src/main/scala/com/yahoo/maha/core/query/presto/PrestoQueryGenerator.scala
panditsurabhi: Checked with a few queries in prod, performance seems same.

core/src/test/scala/com/yahoo/maha/core/query/presto/BasePrestoQueryGeneratorTest.scala
panditsurabhi: For testing different data types

PR #71, opened 2018-01-23
Async more join conditions: query optimisation 

pranavbhole:


pranavbhole:
Working on moving join type to request model

core/src/main/scala/com/yahoo/maha/core/query/JoinType.scala
```diff
@@ -0,0 +1,6 @@
+package com.yahoo.maha.core.query
+
+trait JoinType
```
patelh: sealed

PR #74, opened 2018-01-30
Inner join dims in dim only query type if nonfkpk filter is requested in nondriving dim

pranavbhole:


pranavbhole:
Build is passing

pkdhiman:
👍

PR #77, opened 2018-02-05
Allow multiple fact definitions for the same underlying table

panditsurabhi:


core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -35,6 +35,7 @@ class QueryBuilderContext {
     tableAliasMap.get(name) match {
       case None => {
         val alias = s"${name.split("_").map(_.substring(0, 1)).mkString("")}$aliasIndex"
+        println("Name: "  + name + " Alias: " + alias + " map: " + tableAliasMap)
```
patelh: do u need this?

PR #95, opened 2018-02-23
Fix rowlist null exception

patelh:


pranavbhole:
Looks good to me

pranavbhole:
👍

pranavbhole:
one of the concern is that it will silently skip adding rows on null indexAlias util we add watchdog on this error.

PR #99, opened 2018-02-28
Query pipeline expose request

ryankwagner:


core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -131,13 +136,17 @@ case class DefaultQueryPipeline(queryChain: QueryChain
   }
 }
 
-class QueryPipelineBuilder(queryChain: QueryChain
+class QueryPipelineBuilder(requestModel: RequestModel
+                           , queryChain: QueryChain
```
pranavbhole: RequestModel is already part of QueryContext which is part of QueryChain

PR #122, opened 2018-03-20
RequestModel and RequestResult validators

pranavbhole:


service/src/main/scala/com/yahoo/maha/service/MahaServiceMonitor.scala
```diff
@@ -0,0 +1,23 @@
...
+ */
+trait MahaServiceMonitor {
+  def start(reportingRequest: ReportingRequest): Unit
+  def stop(reportingRequest: ReportingRequest): Unit
+}
+
+object DefaultMahaServiceMonitor extends MahaServiceMonitor {
+  override def start(reportingRequest: ReportingRequest) =  {}
+
+  override def stop() = {}
```
patelh: This needs to be fixed.

PR #125, opened 2018-03-22
Remove Either, Option, Right, Left and only use scala builtins.

ryankwagner:
This removed the custom java functions:
Option
Either
->Left
->Right
As well as associated classes.


pranavbhole:
@ryankwagner  maha-par-request-2 is not yet released , 
you need to add yahoo/maven/maha-par-request-2 as travis component 
also need add config file for that component. 
Finally add publish step for yahoo/maven/maha-par-request-2.

PR #139, opened 2018-03-26
Update Maha Service to depend on Par-Request 2

ryankwagner:


service/pom.xml
```diff
@@ -21,6 +21,11 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.yahoo.maha</groupId>
+            <artifactId>maha-par-request-2</artifactId>
+            <version>${project.version}</version>
+        </dependency>
         <dependency>
```
pranavbhole: exclude maha-par-request from dependency

PR #167, opened 2018-04-06
Drilldown curator

ryankwagner:


service/src/main/scala/com/yahoo/maha/service/curators/DrilldownCurator.scala
```diff
@@ -0,0 +1,221 @@
...
+                       , mahaRequestContext: MahaRequestContext
+                       , mahaService: MahaService
+                       , mahaRequestLogBuilder: CuratorMahaRequestLogBuilder
+                       , curatorConfig: CuratorConfig): ParRequest[CuratorResult] = {
+
+    val registryConfig = mahaService.getMahaServiceConfig.registry(mahaRequestContext.registryName)
+    val parallelServiceExecutor = registryConfig.parallelServiceExecutor
+    val parRequestLabel = "processDrillDownCurator"
+    val firstRequest : ParRequest[CuratorResult] = resultMap(DefaultCurator.name)
+    require(firstRequest.get.isRight, "First par request failed, cannot build the second! ")
```
patelh: This would be a blocking call.  You should just do a flatMap on it.

PR #175, opened 2018-04-11
Fix bunch of stuff

patelh:


api-jersey/src/main/scala/com/yahoo/maha/api/jersey/JsonStreamingOutput.scala
```diff
@@ -58,14 +58,17 @@ case class JsonStreamingOutput(resultList: IndexedSeq[CuratorResult],
         val dimCols : Set[String]  = if(curatorResult.requestModelReference.model.bestCandidates.isDefined) {
           curatorResult.requestModelReference.model.bestCandidates.get.publicFact.dimCols.map(_.alias)
         } else Set.empty
+
+        val includeRowCount: Boolean = curatorResult.requestModelReference.model.includeRowCount && rowCountOption.isDefined
```
patelh: You should remove the && condition since rowCountOption is only defined if it is multi engine.  For single engine, row count is fetched in same query.  You should only use the reportingRequest as a basis for this, not even the model.

api-jersey/src/main/scala/com/yahoo/maha/api/jersey/JsonStreamingOutput.scala
```diff
@@ -169,7 +174,7 @@ case class JsonStreamingOutput(resultList: IndexedSeq[CuratorResult],
           jsonGenerator.writeObject(row.getValue(i))
           i+=1
         }
-        if (rowCountOption.isDefined) {
+        if (includeRowCount) {
```
patelh: This is also wrong, it only needs to be injected if multi engine query hence check on rowCountOption.

PR #179, opened 2018-04-12
CuratorResultPostProcessor with test

pranavbhole:


ryankwagner:
👍

service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
```diff
@@ -113,4 +123,15 @@ case class DefaultCurator(protected val requestModelValidator: CuratorRequestMod
...
   }
 
+  def validateWithTry(curatorResult: CuratorResult): CuratorResult = {
+    try {
+      curatorResultPostProcessor.process(curatorResult)
+    }
+    catch {
+      case e:Exception =>
+        error("Failure in the curatorResultPostProcessor", e)
+        curatorResult
```
patelh: how is this failure captured?

PR #180, opened 2018-04-12
Integrate CuratorConfig into DrillDown Coordinator Pipeline

ryankwagner:


service/src/main/scala/com/yahoo/maha/service/curators/DrilldownConfig.scala
```diff
@@ -2,29 +2,37 @@ package com.yahoo.maha.service.curators
...
+import org.json4s.scalaz.JsonScalaz
 
 /**
   * Parse an input JSON and convert it to a DrilldownConfig object.
   **/
-object DrilldownConfig {
-  val MAXIMUM_ROWS : BigInt = 1000
-  val DEFAULT_ENFORCE_FILTERS : Boolean = false
-  val validCubes : List[String] = List("performance_stats", "user_stats", "student_performance")
+object DrilldownConfig extends CuratorConfig {
```
patelh: this shouldn't extend CuratorConfig

PR #189, opened 2018-04-23
Add presto/hive factories : IN PROGRESS

ryankwagner:


service/src/main/scala/com/yahoo/maha/service/factory/QueryGeneratorFactory.scala
```diff
@@ -92,5 +94,69 @@ class DruidQueryGeneratorFactory extends QueryGeneratorFactory {
...
 
+  override def supportedProperties: List[(String, Boolean)] = List.empty
+}
+
+class HiveQueryGeneratorFactory extends QueryGeneratorFactory {
+  """
+    |{
+    |"partitionColumnRendererClass" : "DefaultPartitionColumnRendererFactory",
+    |"partitionColumnRendererConfig" : [{"key": "value"}],
+    |}
```
pranavbhole: add the config to accept the default registration factory

service/pom.xml
```diff
@@ -54,6 +54,11 @@
             <artifactId>maha-oracle-executor</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.yahoo.maha</groupId>
```
pranavbhole: can you please update maha service example

PR #202, opened 2018-04-30
Drilldownfix

ryankwagner:


service/src/main/scala/com/yahoo/maha/service/curators/DrilldownCurator.scala
```diff
@@ -156,14 +157,80 @@ class DrilldownCurator (override val requestModelValidator: CuratorRequestModelV
...
+      else{
+          selectedRevisionTry.toOption
+      }
+
+      var factFieldsReduced: IndexedSeq[Field] = IndexedSeq.empty
+
+      val bucketSelectedRevision: Int = selectedRevision.get
+
+      factFieldsReduced = factFields.filter (field =>
+        factMap (
```
patelh: why do this lookup for every field?  Just pull it outside the filter...
```
val publicFact = factMap(drilldownConfig.cube, bucketSelectedRevision)
factFieldsReduced = factFields.filter (field =>publicFact.columnsByAlias.contains(field.field))
```

service/src/main/scala/com/yahoo/maha/service/curators/DrilldownCurator.scala
```diff
@@ -156,14 +157,78 @@ class DrilldownCurator (override val requestModelValidator: CuratorRequestModelV
...
+            throw new IllegalArgumentException(t.getMessage)
+        }
+
+
+      }
+      else{
+          selectedRevisionTry.toOption
+      }
+
+      var factFieldsReduced: IndexedSeq[Field] = IndexedSeq.empty
```
patelh: Why do you need a var here?  Why not just do single assignment with val since you only set it once?
```
val factFieldsReduced: IndexedSeq[Field] = factFields.filter (field =>
  pubFact.columnsByAlias.contains (field.field)
)
```

PR #224, opened 2018-05-18
Not In filter throws error for static mapped columns

shek1608:
Currently, when you use a 'Not In' filter for a statically mapped field, it throws an error - _"Unsupported filter operation on statically mapped field"_

* Added the case to support _NotIn Filter_ in _RequestModel.scala_
* Added test
* mvn clean install ✅ 
* updated version using travis/update_snapshot.sh

@pranavbhole Please Review


pranavbhole:
can you please update revision again?

shek1608:
@pranavbhole Have updated revision to 134

PR #229, opened 2018-05-24
Outer-Group-By support for HiveQueryGenerator

panditsurabhi:


core/src/test/scala/com/yahoo/maha/core/query/hive/BaseHiveQueryGeneratorTest.scala
```diff
@@ -336,4 +337,186 @@ trait BaseHiveQueryGeneratorTest
...
+          PublicFactCol("impression_share_rounded", "Impression Share", InBetweenEquality)
+        ),
+        forcedFilters,
+        getMaxDaysWindow, getMaxDaysLookBack
+      )
+  }
+
+
+
+  case object GetIntervalDateRegistration extends UDF {
```
patelh: This all looks specific to our use case, can you make this generic?

PR #233, opened 2018-06-07
write missing lookup dimensions to missingLookup Kafka Topic

pavanab4u:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/InMemoryDBLookupExtractor.java
```diff
@@ -90,6 +93,7 @@ public String apply(@NotNull String val)
             }
 
             if (cacheByteValue == null || cacheByteValue.length == 0) {
+                kafkaManager.handleMissingLookup(extractionNamespace, dimension);
```
patelh: Do you want to do a local cache which has the most recent request missing lookups so you don't do multiple writes for the same key ?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/KafkaManager.java
```diff
@@ -54,13 +59,20 @@
...
                         @Named("kafkaProperties") final Properties kafkaProperties,
                         ProtobufSchemaFactory protobufSchemaFactory) {
         this.kafkaProperties.putAll(kafkaProperties);
         this.namespaceExtractionCacheManager = namespaceExtractionCacheManager;
         this.protobufSchemaFactory = protobufSchemaFactory;
+        final Properties properties = new Properties();
+        properties.putAll(kafkaProperties);
+        properties.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
+        properties.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, ByteArraySerializer.class.getName());
+        this.kafkaProducer = new KafkaProducer<>(properties);
```
patelh: Probably should configure the producer to drop messages if the underlying message cannot be written, we don't want to queue / batch the messages beyond a max value.  A failed write to kafka may just enqueue the message for the send thread and we don't want it to grow unbounded, and we don't want it to block or through exceptions.  It should drop messages if it can't write after queue is full.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/KafkaManager.java
```diff
@@ -313,4 +329,14 @@ public Boolean call() {
             return true;
         }
     }
+
+    public void handleMissingLookup(InMemoryDBExtractionNamespace extractionNamespace, String dimension) {
```
patelh: Can we enable or disable this through configuration ?

PR #246, opened 2018-06-19
Handle HiveDimCol in HiveQueryGenerator

panditsurabhi:


core/src/test/scala/com/yahoo/maha/core/query/hive/BaseHiveQueryGeneratorTest.scala
```diff
@@ -355,6 +355,7 @@ trait BaseHiveQueryGeneratorTest
           , DimCol("start_time", IntType())
           , DimCol("stats_date", DateType("YYYY-MM-dd"))
           , DimCol("show_flag", IntType())
+          , HiveDimCol("stats_hour", StrType())
```
patelh: HiveDimCol is meant to be used for columns which are dim cols in Hive but are not carried over when creating definitions for another engine, e.g. snapshot_ts or shard_id.  In general, I don't expect them to be rendered except for special cases like snapshot_ts.

panditsurabhi: This seems to be the most appropriate way to prevent base fact dim columns from being carried over to the alternate fact. That's because HiveDimCol has EngineRequirement which prevents it from being copied over. This came up while removing stats_hour from Druid for user_stats.

panditsurabhi:
We added discarding set while creating fact with alternate engine to handle the use case.

PR #262, opened 2018-07-02
cache lookup query element and use it before serializing

pavanab4u:


core/src/test/scala/com/yahoo/maha/core/query/druid/DruidQueryGeneratorTest.scala
```diff
@@ -1710,6 +1710,41 @@ class DruidQueryGeneratorTest extends BaseDruidQueryGeneratorTest {
     result should fullyMatch regex json
   }
 
+  /*test("Primary key should be included in if it is present in sort by when druid+oracle") {
```
patelh: Do you need this test if it is commented out?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/MahaRegisteredLookupExtractionFn.java
```diff
@@ -60,6 +69,7 @@ public MahaRegisteredLookupExtractionFn(
         this.valueColumn = valueColumn;
         this.decodeConfig = decodeConfig;
         this.dimensionOverrideMap = dimensionOverrideMap;
+        this.useQueryLevelCache = useQueryLevelCache == null ? false : useQueryLevelCache;
```
patelh: Why not instantiate cache here if the above is set to true?

pavanab4u: I need to lazily instantiate cache. If I eagerly instantiate then maha-core which uses maha-druid-lookups while generating druid query unnecessarily instantiates cache.

PR #268, opened 2018-07-10
Shared DataSource instances across maha service

pranavbhole:


service/src/main/scala/com/yahoo/maha/service/factory/QueryExecutoryFactory.scala
```diff
@@ -146,28 +160,36 @@ class PrestoQueryExecutoryFactory extends QueryExecutoryFactory {
...
-      dataSourceFactoryConfig <- dataSourceFactoryConfigResult
-      dataSourceFactory <-  getFactory[DataSourceFactory](dataSourceFactoryClass, this.closer)
-      dataSource <- dataSourceFactory.fromJson(dataSourceFactoryConfig)
+      dataSourceName <- dataSourceNameResult
+      dataSource <- context.dataSourceMap.get(dataSourceName).successNel[Option[DataSource]].asInstanceOf[MahaServiceConfig.MahaConfigResult[Option[DataSource]]]
       jdbcConnectionFetchSizeOption <- jdbcConnectionFetchSizeOptionResult
     } yield {
       if(jdbcConnectionFetchSizeOption.isDefined) {
-        new JdbcConnection(dataSource, jdbcConnectionFetchSizeOption.get)
+        new JdbcConnection(dataSource.get, jdbcConnectionFetchSizeOption.get)
```
patelh: This is unsafe get on dataSource.  Check if it is defined first.

service/src/main/scala/com/yahoo/maha/service/factory/QueryExecutoryFactory.scala
```diff
@@ -146,28 +160,36 @@ class PrestoQueryExecutoryFactory extends QueryExecutoryFactory {
...
       } else {
-        new JdbcConnection(dataSource)
+        new JdbcConnection(dataSource.get)
```
patelh: Same here, unsafe get

service/src/main/scala/com/yahoo/maha/service/factory/QueryExecutoryFactory.scala
pranavbhole: datasource.get is already validated above if not defined, execution return the MahaServiceError list.

service/src/main/scala/com/yahoo/maha/service/factory/QueryExecutoryFactory.scala
pranavbhole: same here

service/src/main/scala/com/yahoo/maha/service/factory/QueryExecutoryFactory.scala
patelh: Can you add a test for this?

pranavbhole: Added tests for checking unknown datasource

PR #270, opened 2018-07-12
Qgen versioning

panditsurabhi:


core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -260,18 +260,45 @@ object QueryGeneratorHelper {
...
+    require(!isEngineRegistered(engine, version), s"Query generator already defined for engine : $engine and version $version")
+    if (queryGeneratorRegistry.contains(engine)) {
+      queryGeneratorRegistry += (engine -> (Map(version -> qg) ++ queryGeneratorRegistry.get(engine).get))
+    } else {
+      queryGeneratorRegistry += (engine -> Map(version -> qg))
+    }
+  }
+
+  def getGenerator(engine: Engine, version: Option[Version]): Option[QueryGenerator[EngineRequirement]] = {
+    if (queryGeneratorRegistry.get(engine).isDefined) {
```
patelh: Could use for yield instead of if else block
```
for {
  versionGeneratorMap <- queryGeneratorRegistry.get(engine)
  generator <- versionGeneratorMap.get(version.getOrElse(Version.DEFAULT))
} yield generator

File: core/src/main/scala/com/yahoo/maha/core/bucketing/BucketingConfig.scala [hunk 2]
```diff
@@ -81,22 +78,91 @@ class CubeBucketingConfigBuilder {
...
+class DefaultBucketingConfig(cubeBucketingConfigMap:scala.collection.immutable.Map[String,CubeBucketingConfig],
+                             queryGenBucketingConfigMap:scala.collection.immutable.Map[Engine,QueryGenBucketingConfig]) extends BucketingConfig {
   validate()
 
   private[this] def validate(): Unit = {
-    for((cubeName,bucketingConfig)<-bucketingConfigMap) {
+    for((cubeName,bucketingConfig)<-cubeBucketingConfigMap) {
       bucketingConfig.validate(cubeName)
     }
+    for((_,bucketingConfig)<-queryGenBucketingConfigMap) {
```
patelh: Could be rewritten as:
```
queryGenBucketingConfigMap.foreach {
  case (_, bucketingConfig) => bucketingConfig.validate()
}
```

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -260,18 +260,45 @@ object QueryGeneratorHelper {
...
+
 class QueryGeneratorRegistry {
-  private[this] var queryGeneratorRegistry : Map[Engine, QueryGenerator[_]] = Map.empty
 
-  def isEngineRegistered(engine: Engine): Boolean = synchronized {
-    queryGeneratorRegistry.contains(engine)
+  private[this] var queryGeneratorRegistry : Map[Engine, Map[Version , QueryGenerator[_]]] = Map.empty
+
+  def isEngineRegistered(engine: Engine, version: Version = Version.DEFAULT): Boolean = synchronized {
+    queryGeneratorRegistry.contains(engine) && queryGeneratorRegistry.get(engine).get.contains(version)
```
patelh: Could short circuit get if you already know contains is true
```
queryGeneratorRegistry.contains(engine) && queryGeneratorRegistry(engine).contains(version)
```

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -260,18 +260,45 @@ object QueryGeneratorHelper {
...
   }
 
-  def register[U <: QueryGenerator[_]](engine: Engine, qg: U) : Unit = synchronized {
-    require(!queryGeneratorRegistry.contains(engine), s"Query generator already defined for engine : $engine")
-    queryGeneratorRegistry += (engine -> qg)
+  def register[U <: QueryGenerator[_]](engine: Engine, qg: U, version: Version = Version.DEFAULT) : Unit = synchronized {
+    require(!isEngineRegistered(engine, version), s"Query generator already defined for engine : $engine and version $version")
+    if (queryGeneratorRegistry.contains(engine)) {
+      queryGeneratorRegistry += (engine -> (Map(version -> qg) ++ queryGeneratorRegistry.get(engine).get))
```
patelh: Same here, if you already know it contains, you can call apply directly
```
queryGeneratorRegistry += (engine -> (Map(version -> qg) ++ queryGeneratorRegistry(engine)))
```

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -617,17 +625,21 @@ object DefaultQueryPipelineFactory extends Logging {
...
-  private[this] def getDimOnlyQuery(bestDimCandidates: SortedSet[DimensionBundle], requestModel: RequestModel): Query = {
+  private[this] def getDimOnlyQuery(bestDimCandidates: SortedSet[DimensionBundle], requestModel: RequestModel, queryGenVersion: Version): Query = {
     val dimOnlyContextBuilder = QueryContext
       .newQueryContext(DimOnlyQuery, requestModel)
       .addDimTable(bestDimCandidates)
     require(bestDimCandidates.nonEmpty, "Cannot generate dim only query with no best dim candidates!")
     require(queryGeneratorRegistry.isEngineRegistered(bestDimCandidates.head.dim.engine)
       , s"Failed to find query generator for engine : ${bestDimCandidates.head.dim.engine}")
-    queryGeneratorRegistry.getGenerator(bestDimCandidates.head.dim.engine).get.generate(dimOnlyContextBuilder.build())
+    queryGeneratorRegistry.getGenerator(bestDimCandidates.head.dim.engine, Some(queryGenVersion)).get.generate(dimOnlyContextBuilder.build())
```
patelh: Wrap it in Option instead of Some, in case someone passes in null... Option(queryGenVersion)

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -636,24 +648,26 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
 
     require(queryGeneratorRegistry.isEngineRegistered(bestDimCandidates.head.dim.engine)
       , s"Failed to find query generator for engine : ${bestDimCandidates.head.dim.engine}")
-    queryGeneratorRegistry.getGenerator(bestDimCandidates.head.dim.engine).get.generate(dimOnlyContextBuilder.build())
+    queryGeneratorRegistry.getGenerator(bestDimCandidates.head.dim.engine, Some(queryGenVersion)).get.generate(dimOnlyContextBuilder.build())
```
patelh: Same here, use Option instead of Some

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -636,24 +648,26 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
...
-  private[this] def getFactQuery(bestFactCandidate: FactBestCandidate, requestModel: => RequestModel, indexAlias: String): Query = {
+  private[this] def getFactQuery(bestFactCandidate: FactBestCandidate, requestModel: => RequestModel, indexAlias: String, queryGenVersion: Version): Query = {
     val factOnlyContextBuilder = QueryContext
       .newQueryContext(FactOnlyQuery, requestModel)
       .addFactBestCandidate(bestFactCandidate)
       .addIndexAlias(indexAlias)
     require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine)
       , s"Failed to find query generator for engine : ${bestFactCandidate.fact.engine}")
-    queryGeneratorRegistry.getGenerator(bestFactCandidate.fact.engine).get.generate(factOnlyContextBuilder.build())
+    queryGeneratorRegistry.getGenerator(bestFactCandidate.fact.engine, Some(queryGenVersion)).get.generate(factOnlyContextBuilder.build())
```
patelh: Same here, use Option instead of Some

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -665,7 +679,7 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
 
     require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine)
       , s"No query generator registered for engine ${bestFactCandidate.fact.engine} for fact ${bestFactCandidate.fact.name}")
-    queryGeneratorRegistry.getGenerator(bestFactCandidate.fact.engine).get.generate(dimFactContext)
+    queryGeneratorRegistry.getGenerator(bestFactCandidate.fact.engine, Some(queryGenVersion)).get.generate(dimFactContext)
```
patelh: Option instead of Some

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -735,7 +752,7 @@ OuterGroupBy operation has to be applied only in the following cases
 
         require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine)
           , s"No query generator registered for engine ${bestFactCandidate.fact.engine} for fact ${bestFactCandidate.fact.name}")
-        queryGeneratorRegistry.getGenerator(bestFactCandidate.fact.engine).get.generate(factContext)
+        queryGeneratorRegistry.getGenerator(bestFactCandidate.fact.engine, Some(queryGenVersion)).get.generate(factContext)
```
patelh: Option instead of Some

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -260,18 +260,45 @@ object QueryGeneratorHelper {
...
+object V2 extends Version(2)
+
 class QueryGeneratorRegistry {
-  private[this] var queryGeneratorRegistry : Map[Engine, QueryGenerator[_]] = Map.empty
 
-  def isEngineRegistered(engine: Engine): Boolean = synchronized {
-    queryGeneratorRegistry.contains(engine)
+  private[this] var queryGeneratorRegistry : Map[Engine, Map[Version , QueryGenerator[_]]] = Map.empty
+
+  def isEngineRegistered(engine: Engine, version: Version = Version.DEFAULT): Boolean = synchronized {
```
patelh: You should not have default value for version, it should be passed in, otherwise you may end up verifying an engine exists for default version and not the version passed into getGenerator.  You should make this method take same params as getGenerator.

panditsurabhi: We can't know which qgen version to use before the engine is decided. 'isEngineRegistered' is called  in the decision process. I have made it an option and removed default so that callers are forced to pass a value.

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -260,18 +260,43 @@ object QueryGeneratorHelper {
...
+    }
   }
 
-  def register[U <: QueryGenerator[_]](engine: Engine, qg: U) : Unit = synchronized {
-    require(!queryGeneratorRegistry.contains(engine), s"Query generator already defined for engine : $engine")
-    queryGeneratorRegistry += (engine -> qg)
+  def getGenerator(engine: Engine, version: Option[Version]): Option[QueryGenerator[EngineRequirement]] = {
+    for {
+      versionGeneratorMap <- queryGeneratorRegistry.get(engine)
+      generator <- versionGeneratorMap.get(version.getOrElse(Version.DEFAULT)).asInstanceOf[Option[QueryGenerator[EngineRequirement]]]
```
patelh: why the casting?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -617,43 +625,49 @@ object DefaultQueryPipelineFactory extends Logging {
...
 class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGeneratorRegistry, druidMultiQueryEngineList: Seq[Engine] = DefaultQueryPipelineFactory.druidMultiQueryEngineList) extends QueryPipelineFactory with Logging {
 
-  private[this] def getDimOnlyQuery(bestDimCandidates: SortedSet[DimensionBundle], requestModel: RequestModel): Query = {
+  private[this] def getDimOnlyQuery(bestDimCandidates: SortedSet[DimensionBundle], requestModel: RequestModel, queryGenVersion: Version): Query = {
     val dimOnlyContextBuilder = QueryContext
       .newQueryContext(DimOnlyQuery, requestModel)
       .addDimTable(bestDimCandidates)
     require(bestDimCandidates.nonEmpty, "Cannot generate dim only query with no best dim candidates!")
-    require(queryGeneratorRegistry.isEngineRegistered(bestDimCandidates.head.dim.engine)
+    require(queryGeneratorRegistry.isEngineRegistered(bestDimCandidates.head.dim.engine, None)
```
patelh: Why pass in None when you have the queryGenVersion?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -617,43 +625,49 @@ object DefaultQueryPipelineFactory extends Logging {
...
+                                           queryAttributes: QueryAttributes,
+                                           queryGenVersion: Version): Query = {
     val dimOnlyContextBuilder = QueryContext
       .newQueryContext(DimOnlyQuery, requestModel)
       .addDimTable(bestDimCandidates)
       .addIndexAlias(indexAlias)
       .setQueryAttributes(queryAttributes)
 
-    require(queryGeneratorRegistry.isEngineRegistered(bestDimCandidates.head.dim.engine)
+    require(queryGeneratorRegistry.isEngineRegistered(bestDimCandidates.head.dim.engine, None)
```
patelh: Why pass in None when you have the queryGenVersion?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -617,43 +625,49 @@ object DefaultQueryPipelineFactory extends Logging {
...
   }
 
-  private[this] def getFactQuery(bestFactCandidate: FactBestCandidate, requestModel: => RequestModel, indexAlias: String): Query = {
+  private[this] def getFactQuery(bestFactCandidate: FactBestCandidate, requestModel: => RequestModel, indexAlias: String, queryGenVersion: Version): Query = {
     val factOnlyContextBuilder = QueryContext
       .newQueryContext(FactOnlyQuery, requestModel)
       .addFactBestCandidate(bestFactCandidate)
       .addIndexAlias(indexAlias)
-    require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine)
+    require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine, None)
```
patelh: Why pass in None when you have the queryGenVersion?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -663,9 +677,9 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
       .setQueryAttributes(queryAttributes)
       .build()
 
-    require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine)
+    require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine, None)
```
patelh: Why pass in None when you have the queryGenVersion?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -733,9 +750,9 @@ OuterGroupBy operation has to be applied only in the following cases
           .setQueryAttributes(queryAttributes)
           .build()
 
-        require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine)
+        require(queryGeneratorRegistry.isEngineRegistered(bestFactCandidate.fact.engine, None)
```
patelh: Why pass in None when you have the queryGenVersion?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -761,36 +778,45 @@ OuterGroupBy operation has to be applied only in the following cases
...
+    val queryPipelineTryDefault = builder(requestModels._1, queryAttributes, None)._1
     var queryPipelineTryDryRun: Option[Try[QueryPipelineBuilder]] = None
     if (requestModels._2.isDefined) {
-      queryPipelineTryDryRun =  Some(builder(requestModels._2.get, queryAttributes))
+      queryPipelineTryDryRun =  Option(builder(requestModels._2.get, queryAttributes, None)._1)
     }
     (queryPipelineTryDefault, queryPipelineTryDryRun)
   }
 
+  def builder(requestModels: Tuple2[RequestModel, Option[RequestModel]], queryAttributes: QueryAttributes, bucketSelector: BucketSelector): Tuple3[Try[QueryPipelineBuilder], Option[Try[QueryPipelineBuilder]], Option[Try[QueryPipelineBuilder]]] = {
```
patelh: why 3 pipelines?

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -951,49 +980,14 @@ OuterGroupBy operation has to be applied only in the following cases
       val factBestCandidateOption =
         requestModel.bestCandidates.map(_ => findBestFactCandidate(requestModel, forceDisqualifyEngine, dimEngines))
       val dimensionCandidates = findBestDimCandidates(requestModel, factBestCandidateOption.map(_.fact.engine).getOrElse(OracleEngine), dimensionCandidatesMapping).filter(db => {
-        val queryGenerator = queryGeneratorRegistry.getGenerator(db.dim.engine)
+        val queryGenerator = queryGeneratorRegistry.getGenerator(db.dim.engine, None)
```
patelh: This doens't look right.  Need to verify the engine constraints on generator and you want to do it on the version of generator used.

panditsurabhi: I realized that, but at this point, engine is not finalized so we cannot select query generator buckets, and hence don't know the version. 

We can select buckets (and versions) for all engines at the beginning but then bucket percentages will not mean what they should, because all requests will result in random number generation for all engines, whether that engine is used or not.

Another option is to move validateEngineConstraints() to the Engine objects (HiveEngine, OracleEngine etc), because it seems to be engine specific, not query generator specific.

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
panditsurabhi: Wasn't compiling - "Expression of type Option[QueryGenerator[_]]] doesn't confirm to Option[QueryGenerator[EngineRequirement]]]", fixed it now

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -260,18 +260,59 @@ object QueryGeneratorHelper {
...
+  def number: Int
+}
+case class Version private(number: Int) extends VersionNumber
+object Version {
+  val v0: Version = new Version(0)
+  val v1: Version = new Version(1)
+  val v2: Version = new Version(2)
+  val DEFAULT = v0
+
+  def apply(number: Int) : Version = {
```
patelh: can't we make this private?

core/src/main/scala/com/yahoo/maha/core/bucketing/BucketSelector.scala
```diff
@@ -75,13 +107,35 @@ class BucketSelector(registry: Registry, bucketingConfig: BucketingConfig) exten
     }
   }
 
+  private def selectVersion(queryGenConfig: Option[QueryGenBucketingConfig], requestParams: BucketParams): Version = {
+    if (requestParams.userInfo.isInternal) {
+      Version.apply(queryGenConfig.get.internalDistribution.sample())
```
patelh: We should treat Version instances (v0, v1, v2) as singletons.  Add new method e.g. Version.from(int): Option[Version] if it exists.  You can create a map to look up the versions in Version, e.g. val versions: Map[Int, Version] = List(v0, v1, v2).map(v -> (v.number, v)).toMap

PR #291, opened 2018-08-07
Add codecs and kafka serdes

patelh:


db/src/main/java/com/yahoo/maha/data/ByteBufferOutputStream.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public ByteBufferOutputStream() {
+    }
+
+    public void set(ByteBuffer buffer) {
+        this.buffer = buffer;
+    }
+
+    public void write(byte[] b) throws IOException {
+        if(b.length > this.buffer.capacity() - this.buffer.position()) {
```
panditsurabhi: null check for b?

db/src/main/java/com/yahoo/maha/data/Compressor.java
```diff
@@ -0,0 +1,36 @@
...
+    int decompress(byte[] bytes, byte[] outputBuffer) throws IOException;
+    void decompressBB(ByteBuffer byteBuffer, ByteBuffer outputBuffer) throws IOException;
+
+    enum Codec {
+        GZIP(1,1), LZ4(2,2), LZ4HC(3,3), NONE(4,4);
+
+        private final int index;
+        private final int value;
+
+        private Codec(int index, int value) {
```
panditsurabhi: Do we need index/value? Didn't see their use anywhere.

db/src/main/java/com/yahoo/maha/data/PassThroughCodec.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public void decompressToBB(byte[] bytes, ByteBuffer outputBuffer) throws IOException {
+        outputBuffer.put(bytes);
+    }
+
+    @Override
+    public int decompress(byte[] bytes, byte[] outputBuffer) throws IOException {
+        System.arraycopy(bytes, 0, outputBuffer, 0, bytes.length);
```
panditsurabhi: arraycopy can result in RuntimeExceptions, like IndexOOB. If we are declaring IOException, then may be we should wrap in try-catch and throw IOException, or remove the 'throws' clause.

db/src/test/java/com/yahoo/maha/data/StringEventBatchDecoder.java
```diff
@@ -0,0 +1,96 @@
...
+    private final LazyInitializer<ByteBuffer> lazyInputByteBuffer;
+    private final LazyInitializer<ByteBuffer> lazyOutputByteBuffer;
+    private final LazyInitializer<byte[]> lazyOutputByteArray;
+    private final Compressor compressor;
+
+    StringEventBatchDecoder(Compressor.Codec codec) {
+        this(codec, 1);
+    }
+    StringEventBatchDecoder(Compressor.Codec codec, int bufferMb) {
+        bufferSize = bufferMb * 1024 * 1024;
```
panditsurabhi: FileUtils has some readable constants for MB, GB etc https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/FileUtils.html

db/src/main/java/com/yahoo/maha/data/Compressor.java
patelh: index is there if we want to write the codec version when serializing

db/src/main/java/com/yahoo/maha/data/PassThroughCodec.java
patelh: The throws is tied to the Compressor interface.  This is just one implementation which I expect won't be used much if at all.

request-log/src/main/scala/com/yahoo/maha/log/StringEventBatchSerde.scala
```diff
@@ -0,0 +1,137 @@
...
+        (new Output(1024 * 1024 * DEFAULT_BUFFER_MB), new Array[Byte](1024 * 1024 * DEFAULT_BUFFER_MB))
+    }
+    output = configuredOutput
+    encodeBuffer = configuredBuffer
+    logger.warn("Using maha compression codec : {} , buffer bytes = {}"
+      , compressor.codec().name(), encodeBuffer.length)
+  }
+
+  def serialize(topic: String, data: StringEventBatch): Array[Byte] = {
+    kryo.writeObject(output, data)
```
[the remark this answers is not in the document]
patelh: We don't expect this to be used across threads.  The consumer itself is not thread safe so the user of the consumer needs to worry about thread safety.

PR #292, opened 2018-08-07
Add DynamicMahaServiceConfig

panditsurabhi:


service/src/test/scala/com/yahoo/maha/service/factory/MahaServiceTest.scala
```diff
@@ -550,4 +552,247 @@ class MahaServiceTest extends BaseFactoryTest {
...
+         |         }
+         |      }
+         |   }
+         |}""".stripMargin
+
+    //println(jsonString)
+    val json = parse(jsonString)
+    val dynamicObjects = DynamicMahaServiceConfig.findDynamicProperties(json, Map("druidExec" -> new Object, "erBucket" -> new Object))
+    println(dynamicObjects)
+    DynamicMahaServiceConfig.fromJson(jsonString.getBytes(StandardCharsets.UTF_8))
```
pranavbhole: no asserts?

service/src/main/scala/com/yahoo/maha/service/MahaService.scala
```diff
@@ -657,4 +662,139 @@ object MahaServiceConfig {
     result
   }
 
-}
\ No newline at end of file
+}
+
+
+case class DynamicMahaServiceConfig(dynamicProperties: Map[String, DynamicPropertyInfo], context: MahaServiceConfigContext, registry: Map[String, RegistryConfig], mahaRequestLogWriter: MahaRequestLogWriter, curatorMap: Map[String, Curator])
```
pranavbhole: can you more test to test that how dynamic configs are applied in the case of existing pending execution ?

pranavbhole: and also if dynamic config is failed to apply, what happens, test should include all cases

service/src/main/scala/com/yahoo/maha/service/MahaService.scala
```diff
@@ -657,4 +662,139 @@ object MahaServiceConfig {
...
+
+  def mergeMaps(maps: Map[String, Object]*): mutable.Map[String, Object] = {
+    val mergedMap = new scala.collection.mutable.HashMap[String, Object]()
+    maps.foreach(map => {
+      map.foreach(f => mergedMap.put(f._1, f._2))
+    })
+    mergedMap
+  }
+
+  def findDynamicProperties(json: JValue, objectNameMap: Map[String, Object]): Map[String, DynamicPropertyInfo] = {
```
pranavbhole: objectNameMap has String-> Object, we should use the base trait(can be name as MscObject)  instead of generic object to enforce the implementation in registry configs.

service/src/main/scala/com/yahoo/maha/service/MahaService.scala
panditsurabhi: This is just the creation of dynamic config, not its use. Will add more logic and tests incrementally.

service/src/main/scala/com/yahoo/maha/service/MahaService.scala
panditsurabhi: Can't do that right now because ParallelServiceExecutor is in par-request-2 module which doesn't depend on any other maha module (like core). Will need some refactoring to do it cleanly.

PR #295, opened 2018-08-08
Maha worker reporting proto

pranavbhole:
@panditsurabhi


worker/src/main/proto/MahaCustomReportRequest.proto
```diff
@@ -0,0 +1,64 @@
...
+    optional bool dryrun = 7;
+    optional QueueType queueType = 8;
+    optional string userId = 9;
+    optional bool dualWrite = 10;
+    optional string registry_name = 11;
+}
+
+enum DeliveryMethod {
+    EMAIL = 0;
+    MOBSTOR_AND_API = 1;
```
panditsurabhi: Let's rename MOBSTOR_AND_API to CONTENT_STORE or something like that

worker/src/main/proto/MahaCustomReportRequest.proto
```diff
@@ -0,0 +1,64 @@
...
+    JSON = 1;
+}
+
+enum ReportType {
+    AD_HOC = 0;
+    SCHEDULED = 1;
+}
+
+enum QueueType {
+    CMS = 0;
```
panditsurabhi: Let's rename CMS to PULSAR

worker/src/main/proto/MahaCustomReportRequest.proto
pranavbhole: done

worker/src/main/proto/MahaCustomReportRequest.proto
pranavbhole: done

PR #297, opened 2018-08-09
Druid Limit Spec Update for sorting columns

ashwintumma23:
@pavanab4u @pranavbhole @patelh : For Union View queries, LimitSpec flow gets evaluated twice, hence sortBy columns try to get added twice, and Druid query gen fails with error: `Unknown column in order clause`. Fixing this by checking the aggList and postAggList values, which are empty in one case.


yahoocla:

PR #301, opened 2018-08-15
Fix total rows

patelh:


pranavbhole:
need to run the query on box as Count(*) OVER() is not supported in h2

PR #307, opened 2018-08-20
Add method to generate queryPipeline from bucketSelector

panditsurabhi:


service/src/main/scala/com/yahoo/maha/service/MahaService.scala
```diff
@@ -238,6 +249,17 @@ case class DefaultMahaService(config: MahaServiceConfig) extends MahaService wit
     queryPipelineTry
   }
 
+  def generateQueryPipelinesUsingBucketSelector(registryName: String,
```
pranavbhole: you can use this method everywhere while generating query pipeline

pranavbhole: in maha service

PR #310, opened 2018-08-23
Initial implementation of conditionalForcedFilters

raghukumark:


core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
```diff
@@ -1607,6 +1609,7 @@ case class PublicFactTable private[fact](name: String
                                          , factCols: Set[PublicFactColumn]
                                          , facts: Map[String, Fact]
                                          , forcedFilters: Set[ForcedFilter]
+                                         , conditionalForcedFilter: Option[ConditionalForcedFilter]
```
pranavbhole: validate the conditional filter field names against fact

PR #332, opened 2018-09-12
Require values to be within column length

ritvikjaiswal:


core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -814,6 +814,19 @@ object RequestModel extends Logging {
                         val pubCol = publicDim.columnsByAliasMap(filter.field)
                         require(pubCol.filters.contains(filter.operator),
                           s"Unsupported filter operation : dimension=${publicDim.name}, col=${filter.field}, operation=${filter.operator}, expected=${pubCol.filters}")
+                        require(
+                          publicDim.nameToDataTypeMap(pubCol.name) match {
```
pranavbhole: this is validation just dimension filters, how about the fact filters ?

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -1115,6 +1119,21 @@ object RequestModel extends Logging {
     }
   }
 
+  def validateLengthForFilterValue(publicFactOption: Option[PublicFact], publicDimOption: Option[PublicDimension], filter: Filter): Boolean = {
+    val pubCol = if (publicFactOption.isDefined) publicFactOption.get.columnsByAliasMap(filter.field) else publicDimOption.get.columnsByAliasMap(filter.field)
```
pranavbhole: can you combine these two statements? and also check the publicDim option is not empty.

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -1115,6 +1119,21 @@ object RequestModel extends Logging {
...
+    val dataType = if (publicFactOption.isDefined) publicFactOption.get.dataTypeForAlias(pubCol.alias) else publicDimOption.get.nameToDataTypeMap(pubCol.name)
+    dataType match {
+      case StrType(length, _, _) => filter match {
+        case InFilter(_, values, _, _) if length == 0 && values.forall(_.length <= MAX_ALLOWED_STR_LEN) || values.forall(_.length <= length) => true
+        case NotInFilter(_, values, _, _) if length == 0 && values.forall(_.length <= MAX_ALLOWED_STR_LEN) || values.forall(_.length <= length) => true
+        case EqualityFilter(_, value, _, _) if length == 0 && value.length <= MAX_ALLOWED_STR_LEN || value.length <= length => true
+        case NotEqualToFilter(_, value, _, _) if length == 0 && value.length <= MAX_ALLOWED_STR_LEN || value.length <= length => true
+        case LikeFilter(_, value, _, _) if length == 0 && value.length <= MAX_ALLOWED_STR_LEN || value.length <= length => true
+        case _ => false
```
pranavbhole: please include IsNullFilter and other filters, ideally you should have all the filters here otherwise its going to return false for rest of missing and error will be misleading.

pranavbhole:
Add test cases covering all filters

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -1115,6 +1119,40 @@ object RequestModel extends Logging {
...
+        case EqualityFilter(_, value, _, _) =>
+          if (length == 0 && value.length <= MAX_ALLOWED_STR_LEN || value.length <= length) true else false
+        case NotEqualToFilter(_, value, _, _) =>
+          if (length == 0 && value.length <= MAX_ALLOWED_STR_LEN || value.length <= length) true else false
+        case LikeFilter(_, value, _, _) =>
+          if (length == 0 && value.length <= MAX_ALLOWED_STR_LEN || value.length <= length) true else false
+        case BetweenFilter(_, from, to) =>
+          if (from.length <= MAX_ALLOWED_STR_LEN && to.length <= MAX_ALLOWED_STR_LEN) true else false
+        case IsNullFilter(_, _, _) | IsNotNullFilter(_, _, _) | PushDownFilter(_) | OuterFilter(_) | OrFliter(_) => true
+        case _ => throw new Exception(s"Unhandled FilterOperation.")
```
pranavbhole: print the unhandled operation in the exception

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -1115,6 +1119,40 @@ object RequestModel extends Logging {
...
+      }
+      case _ => filter match {
```
pranavbhole: this code is redundant, we dont want to check the length on other data types, int/ doubles are already checked by the json parser. you can remove this case _ and always return true in this case

core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
```diff
@@ -5372,5 +5372,30 @@ class RequestModelTest extends FunSuite with Matchers {
     assert(model.factFilters.find(_.field === "Device ID").get.asInstanceOf[NotInFilter].values === List("3"))
   }
 
+  test("test validateLengthForFilterValue for publicDim") {
+    val publicDim = defaultRegistry.getDimension("campaign")
+    val field = "Campaign Name"
+    val values = List("abc", "def", "someNameLongerThanColumnLength")
+    val value = "someNameLongerThanColumnLength"
```
pranavbhole: add  a test case with str type length 0

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -708,6 +711,7 @@ object RequestModel extends Logging {
               val pubCol = publicFact.columnsByAliasMap(filter.field)
               require(pubCol.filters.contains(filter.operator),
                 s"Unsupported filter operation : cube=${publicFact.name}, col=${filter.field}, operation=${filter.operator}")
+              require(validateLengthForFilterValue(publicFact, filter), s"Value for ${filter.field} exceeds max length.")
```
pranavbhole: it would be great if you can include expected and actual length

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -711,7 +711,7 @@ object RequestModel extends Logging {
               val pubCol = publicFact.columnsByAliasMap(filter.field)
               require(pubCol.filters.contains(filter.operator),
                 s"Unsupported filter operation : cube=${publicFact.name}, col=${filter.field}, operation=${filter.operator}")
-              require(validateLengthForFilterValue(publicFact, filter), s"Value for ${filter.field} exceeds max length.")
+              require(validateLengthForFilterValue(publicFact, filter)._1, s"Value for ${filter.field} exceeds max length of ${validateLengthForFilterValue(publicFact, filter)._2} characters.")
```
pranavbhole: lets not make 2 calls, you can use extra val to avoid it

PR #367, opened 2018-10-17
Update tests, add archaius callbacks

panditsurabhi:


service/src/main/scala/com/yahoo/maha/service/factory/Factory.scala
```diff
@@ -189,6 +191,21 @@ object DefaultBucketingConfigFactory {
                         , userWhiteList: List[UserRevisionConfig])
   case class CombinedBucketingConfig(cubeBucketingConfigMapList: List[CubeConfig],
                                      queryGenBucketingConfigList: List[QueryGenConfig])
+
+
+  class DBConfigurationSource extends PolledConfigurationSource {
```
patelh: is this a test class?

PR #368, opened 2018-10-18
JavaScript Filter

sapphirew:


core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -141,6 +143,15 @@ case class LessThanFilter(field: String, value: String
   override def canBeHighCardinalityFilter: Boolean = true
 }
 
+case class JavaScriptFilter(field: String, function: String
+                          , override val isForceFilter: Boolean = false
+                          , override val isOverridable: Boolean = false
+                         ) extends ForcedFilter {
+  override def operator = EqualityFilterOperation
```
pavanab4u: operator should not be EqualityFilterOperation. Perhaps, you can have NoopFilterOperator.

PR #378, opened 2018-11-26
Adding Null check for Lookup + reverting to druid 0.11.0

shek1608:
* Added check
* Added test
* Reverted from Druid 0.12.2 to 0.11.0
* mvn clean install ✅


patelh:
Was this a bug fix?  Could you have upgrade to 0.12.3 instead to fix?

pavanab4u:
> Was this a bug fix? Could you have upgrade to 0.12.3 instead to fix?

@patelh this was a bug fix. We've not upgraded Druid cluster to 0.12.* yet. Hence had to revert to 0.11.0 to push this bug fix. We will upgrade to 0.12.3 later.

patelh:
Yikes, surprised there was a client side bug introduced in 0.12.x

PR #379, opened 2018-11-29
adding case for decode_dim in getDruidFilter

mcjyang:


yahoocla:

core/src/test/scala/com/yahoo/maha/core/query/druid/DruidQueryGeneratorTest.scala
```diff
@@ -1604,6 +1603,38 @@ class DruidQueryGeneratorTest extends BaseDruidQueryGeneratorTest {
...
+  test("dimension filter extraction function for decode dim") {
+    val jsonString =
+      s"""{
+                          "cube": "k_stats",
+                          "selectFields": [
+                            {"field": "Advertiser ID"}
+                          ],
+                          "filterExpressions": [
+                            {"field": "Day", "operator": "between", "from": "$fromDate", "to": "$toDate"},
+                            {"field": "Derived Pricing Type", "operator": "in", "values": ["CPV"]}
```
pranavbhole: can you add tests for value which is mapped to 2 or more int values?

PR #385, opened 2018-12-14
Fix Total Rows to remove erroneous fact join

ryankwagner:


patelh:
Hard to see what is being fixed without any tests...

ryankwagner:
totalRows was returning a fact join on non-pk dim filters.  This removes the fact join that it was forcing in RequestModel L615.  I'll add in a test to demonstrate the functionality.

service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
```diff
@@ -257,6 +257,7 @@ case class RowCountCurator(protected val requestModelValidator: CuratorRequestMo
                 //remove all fields except primary key
                 //remove all sorts
                 val primaryKeyAliasFields = sourcePipeline.bestDimCandidates.map(dim => Field(dim.publicDim.primaryKeyByAlias, None, None)).toIndexedSeq
+                val primaryFilteredExpressions = sourcePipeline.requestModel.reportingRequest.filterExpressions.filter(filter => primaryKeyAliasFields.contains(Field(filter.field, None, None)))
```
patelh: I don't think you want to remove all non-pk filters here.  Doesn't make sense given you want to have total rows reflect the number of rows with all filters applied.

ryankwagner: This is true.  I am considering other ways to approach this issue.

patelh: Try seeing why the unexpected join is constructed in request model.

ryankwagner:
Removing this PR, as the removal of filtering is more than what we want.

PR #388, opened 2019-01-11
Demonstrate issue with Total Row query to solve in OracleQueryGenerator.

ryankwagner:


ryankwagner:
Currently only an illustration test PR, proposed solution coming next week.

core/src/main/scala/com/yahoo/maha/core/query/oracle/OracleQueryGenerator.scala
```diff
@@ -234,15 +234,19 @@ b. Dim Driven
       val dimSelect = dimSelectSet.mkString(", ")
       val dimWhere = generateWhereClause(dimBundle, subqueryBundles)
 
+      val isTotalRowRequest = isDimOnly && requestModel.isDimDriven && requestModel.maxRows == 1 && requestModel.includeRowCount
```
patelh: This looks like special casing, the query generator should not know if it is a total rows request (e.g. requestModel.maxRows == 1).  The request model should construct correctly and query pipeline should execute correctly, the generator should not care.  Did you take a look at what Pranav suggested in your previous PR?

ryankwagner: I'm moving the change to QueryPipeline.  The SyncSingleEngine case is the only currently affected, but I am testing if I can insert the logic straight into the queryPipeline builder without harming other queries.

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -949,7 +949,16 @@ OuterGroupBy operation has to be applied only in the following cases
...
         builder
       } else {
-        val query = getDimOnlyQuery(bestDimCandidates, requestModel, queryGenVersion)
+        val isTotalRowQueryWithOneReturnedRow =
+          requestModel.isDimDriven && requestModel.includeRowCount && requestModel.reportingRequest.rowsPerPage == 1 && factBestCandidateOption.isEmpty
+
+        var totalRowRequestModel = requestModel
+
+        if(isTotalRowQueryWithOneReturnedRow)
+          totalRowRequestModel = requestModel.copy(dimSortByMap = Map.empty, factSortByMap = Map.empty, hasDimSortBy = false, hasFactSortBy = false)
```
patelh: If all you are doing is clearing out the sorting for dim and fact, you should do this in the curator.  Not here.  If you want to change how the query is executed by the pipeline, e.g. single vs multi-engine then you should do it here.

service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
```diff
@@ -114,68 +207,40 @@ case class DefaultCurator(protected val requestModelValidator: CuratorRequestMod
...
     } else {
       try {
-        val requestModelResult = requestModelResultTry.get
+        val requestModelResult: RequestModelResult = requestModelResultEither.right.get
         requestModelValidator.validate(mahaRequestContext, requestModelResult)
         val parRequestResult: ParRequestResult = mahaService.executeRequestModelResult(mahaRequestContext.registryName
           , requestModelResult, mahaRequestLogBuilder)
 
+        val (unsortedRequestModelResult, unsortedParRequestResult) =
+          generateUnsortedResults(mahaRequestContext, mahaService, mahaRequestLogBuilder, curatorConfig, requestModelResult, parRequestResult, parRequestResult.queryPipeline.get)
```
pranavbhole: we are already executing the requestModelResult at line no 218? 
and generateUnsortedResults executing it again with constraints ? 

can you avoid double execution, looks like not right.

ryankwagner: Per conversation, I will be converting this change into 2 PRs.  First will be only row count change, second will be only code refactor.  Also, yes this is a double execution that is unnecessary.

PR #396, opened 2019-01-22
Add optimized row count logic into Curator.

ryankwagner:
Check if the query is dimension only, dimension driven, includes row count, and has max rows of 1.  If so, remove ordering.

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
@@ -134,7 +134,8 @@ case class DefaultCurator(protected val requestModelValidator: CuratorRequestMod
             requestModelResult.model.includeRowCount &&
             requestModelResult.model.isDimDriven &&
             requestModelResult.model.maxRows == 1 &&
-            requestModelResult.model.bestCandidates.isEmpty
+            requestModelResult.model.bestCandidates.isEmpty &&
+            requestModelResult.model.dimensionsCandidates.nonEmpty
 
         if(isDimOnlyTotalRowQuery){
           val requestWithoutOrdering = mahaRequestContext.reportingRequest.copy(sortBy = IndexedSeq.empty)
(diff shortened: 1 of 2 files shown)


service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
```diff
@@ -127,8 +127,35 @@ case class DefaultCurator(protected val requestModelValidator: CuratorRequestMod
...
       try {
-        val requestModelResult = requestModelResultTry.get
+        var requestModelResult = requestModelResultTry.get
         requestModelValidator.validate(mahaRequestContext, requestModelResult)
+
+        val isDimOnlyTotalRowQuery: Boolean =
+            requestModelResult.model.includeRowCount &&
+            requestModelResult.model.isDimDriven &&
+            requestModelResult.model.maxRows == 1 &&
+            requestModelResult.model.bestCandidates.isEmpty
```
pranavbhole: it is assumed but its good to add dim candidates non empty

pranavbhole: And also curators are empty

service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
```diff
@@ -127,8 +127,36 @@ case class DefaultCurator(protected val requestModelValidator: CuratorRequestMod
       )
     } else {
       try {
-        val requestModelResult = requestModelResultTry.get
+        var requestModelResult = requestModelResultTry.get
         requestModelValidator.validate(mahaRequestContext, requestModelResult)
+
+        val isDimOnlyTotalRowQuery: Boolean =
```
patelh: If we are changing the default curator, to special case a request which we only care about total rows, and the user is already passing in row count of 1, why can't they just pass empty ordering clause ?

ryankwagner: You're right, the user was passing a problematic query & we shouldn't be updating policy to account for their actions here.  Steps have been taken on that end, and the query is fixed on their end.  Thanks for the input!  Going to close the related PRs for this.

PR #399, opened 2019-01-27
Add Field Equality Filter operation with service example test

ryankwagner:
Currently does not function on Metric fields - only dimensions.

Service test illustrates a selection of top student from each class in a single request, rather than subsequent query generation on top student ID.

Functionality is equivalent on Oracle/Hive/Presto to: WHERE col_a = col_b
On Druid, this is columnComparison filter noted here:
http://druid.io/docs/latest/querying/filters.html


ryankwagner:
Pending work is adding a flag to Filter to indicate if the present filter is multi-column.
This will be used to determine is multiple fields need to be rendered & utilized in all engines.  Presently, one could attempt Filter casting to determine this & then force operations on a specific filter type.  If a future filter is added though, the work would be duplicated.  A set flag circumvents this issue.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -943,8 +1023,13 @@ object FilterSql {
     filter match {
       case PushDownFilter(f) => 
         renderFilter(f, aliasToNameMapFull, columnsByNameMap, engine, literalMapper, expandedExpression)
+      case FieldEqualityFilter(f,g, _, _) =>
+        require(aliasToNameMapFull.contains(g))
```
patelh: We don't want to do validation like this at render time.  We should do this validation before we do all the hard work to get this far.  Ideally in request model generation.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -772,6 +839,19 @@ object FilterDruid {
       case f @ EqualityFilter(alias, value, _, _) => {
         getDruidFilter(grainOption, column, columnAlias, value, columnsByNameMap)
       }
+      case f @ FieldEqualityFilter(field, compareTo, _, _) => {
+        require(aliasToNameMapFull.contains(field) && aliasToNameMapFull.contains(compareTo), s"A requested field is invalid!  Tried: $field & $compareTo")
```
patelh: Same, we don't want to do validation here.  It should be done in request model.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -219,6 +234,17 @@ sealed trait FilterRenderer[T, O] {
              grainOption: Option[Grain]): O
 }
 
+sealed trait FieldFilterRenderer[T, O] {
```
patelh: You need access to the other column, maybe a better way to do this would be to modify the FilterRender trait itself.  E.g.
```
  def render(name: String,
             filter: T,
             literalMapper: LiteralMapper,
             column: Column,
             aliasToNameMapFull: Map[String, String],
             engine: Engine,
             grainOption: Option[Grain]): O
```
Then use this to get the other column:
```
val otherColumn: Column = column.columnContext.getColumnByName(aliasToNameMapFull(filter.compareTo))
```
And you should validate the other column is exists why before you get the render, ideally when creating request model.

patelh:
+1 on adding this feature

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
ryankwagner: issue here is mostly concerned with not having a computed second alias in the case of two metrics being passed.  The name passed into this render function may be "ad_id", or it may be "SUM(IMPRESSIONS)" and I wanted it to be generic enough to handle both.  Metric handling coming soon.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
ryankwagner: With this and the prior suggestion, the first piece of validation (original way) is being done by way of try-catch in the code block here.  I could attempt to move this validation into the requestModel completely.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
patelh: Are you saying the user query would have "SUM(IMPRESSIONS)" or that it would have the field name "Impressions" and the render would need to render it as "SUM(IMPRESSIONS)" ?

ryankwagner: By the time the query calls render in the generator, SUM(IMPRESSIONS) has been created by this inside the respective generator:
val exp = column match {
              case FactCol(_, dt, cc, rollup, _, annotations, _) =>
                s"""${renderRollupExpression(x.name, rollup)}"""
              case PrestoDerFactCol(_, _, dt, cc, de, annotations, rollup, _) =>
                s"""${renderRollupExpression(de.render(x.name, Map.empty), rollup)}"""
              case any =>
                throw new UnsupportedOperationException(s"Found non fact column : $any")
            }

patelh: in that case, maybe we should pass in a map of Alias to Rendered SQL?

ryankwagner: I am now passing a map from alias to (name, rendered sql) so that all rendering functions can use which they need.

ryankwagner: This replaces original name being passed so that there will be less redundant code.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -772,6 +813,18 @@ object FilterDruid {
       case f @ EqualityFilter(alias, value, _, _) => {
         getDruidFilter(grainOption, column, columnAlias, value, columnsByNameMap)
       }
+      case f @ FieldEqualityFilter(field, compareTo, _, _) => {
+        val firstAlias = aliasToNameMapFull(field)
+        val otherAlias = aliasToNameMapFull(compareTo)
+        val defaultDimSpecs: List[DimensionSpec] = (List(firstAlias, otherAlias)).map(
+          (fieldToCompare: String) => {
+            require(columnsByNameMap.contains(fieldToCompare), s"Queried comparison field $fieldToCompare does not exist!")
```
patelh: I thought you moved validation to request model?

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -925,8 +980,10 @@ object FilterSql {
                    engine: Engine,
                    literalMapper: LiteralMapper,
                    expandedExpression: Option[String] = None,
+                   expandedSecondaryExpression: Option[String] = None,
```
patelh: don't you mean to have the aliasToRenderedSqlMap passed in here?

ryankwagner:
All filter renderers in FilterOperation have been updated with the new behavior for consistency.

pranavbhole:
Can you please add the filter validation in the request model to fail fast on invalid names?

pranavbhole:
Thank you for adding request model validation, can you also add tests of request Filter like 
Campaign ID == "Unknown Field" and see if you get 400 back with proper error message.

PR #400, opened 2019-01-29
fix date of week issue

kevinchen3516:
For getting week of year, it should use "xxxx" which indicate week year.

https://www.joda.org/joda-time/apidocs/org/joda/time/format/DateTimeFormat.html


yahoocla:

PR #404, opened 2019-01-31
Simplify pipeline filter render with function call.

ryankwagner:


ryankwagner:
This takes the map population being done via if-else & shrinks it into a function call, to make filter generation easier to follow & read.  This should make addition of OR filter & others simpler to implement.

Issue #392, opened 2019-01-18
OR filter support for all engines


pranavbhole:

ryankwagner:
OrFilter concept query:
`"fields": [
  {"field": "Homeroom Teacher Name"},
  {"field": "Student ID"},
  {"field": "Class ID"},
  {"field": "Class Name"},
  {"field": "Grade"}
]
,"filters": [
{"operator": "=" "fieldName": "Homeroom Teacher Name", "value": "Ms. Olsen"},
  {"operator": "OR" "filters" [
      {"operator": "=", "fieldName": "Student ID", "value": "12345"},
      {"operator": ">", "fieldName": "Grade", "value": "90"},
      {"operator": "<", "fieldName": "Number of Classes", "value": "4"}
    ]
]`

pranavbhole:
Good use case, just few additions to design. CompoundFilterOperation or CombinedFilterOperation filter list should support all the existing filters. Input Reporting Request looks good to me.

ryankwagner:
At the moment, the design looks like it should be able to support all existing, as well as nesting on itself for future usage that is to say: `( = OR = OR ( = AND =) OR =)`

As well as any existing FilterOperations.  It should be entirely agnostic of what sort of FilterOperation passes into it.

patelh:
The current implementation assumes all items in the filter list are AND'd.  We would want to keep the existing functionality intact.  However, the user has the option to use logical operators as necessary (AND, OR, NOT).

Here are some scenarios we should cover:

1. FilterA AND FilterB AND FilterC
    - supported using existing behavior of providing a list of filters
    - supported using an AND logical operator and providing the list of filters to the operator
2. FilterA OR FilterB AND FilterC
    - supported using an OR logical operator with the list of filters being FilterA and an AND logical operator, e.g. OrFilter(List(FilterA, AndFilter(List(FilterB, FilterC))))
3. FilterA AND (FilterB OR FilterC)
    - supported using an AND logical operator with the list of filters being FilterA and an OR logical operator, e.g. AndFilter(List(FilterA, OrFilter(List(FilterB, FilterC))))
4. FilterA AND (FilterB OR (NOT FilterC))
    - supported using an AND logical operator with the list of filters being FilterA and an OR logical operator, e.g. AndFilter(List(FilterA, OrFilter(List(FilterB, NotFilter(FilterC)))))

ryankwagner:
[...]

PR #415, opened 2019-02-13
Add OrFilter operations. (Issue 392)

ryankwagner:
Per issue [392](https://github.com/yahoo/maha/issues/392)

Currently:
- Adds OrFilter Operation without losing access to Druid OrFilter from previous work.

* This is a preparatory step in adding OrFilter Support.
* The next step (on the following PR) will be refactoring RequestModel & QueryGenerator code to make generic the handling of FilterOperations so that in the future new filters can be quickly added.
* After that, helper functions will be created to create lists/sets in place of existing Strings and filter.fields
* Finally, OrFilter support can be added for all Engines.


core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -1296,33 +1350,57 @@ object Filter extends Logging {
...
+  def andFilter(andFilter: AndFilter) : JsonScalaz.Result[Boolean] = {
+    import _root_.scalaz.syntax.validation._
+    if(andFilter.filters.isEmpty) {
+      Fail.apply(andFilter.field, s"filter cannot have empty list")
+    } else {
+      true.successNel
+    }
+  }
+
+  implicit class JValueOps(value: JValue) {
```
patelh: Hm, why do you need this?

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -221,11 +217,27 @@ case class IsNotNullFilter(field: String
...
+}
+
+case class AndFilter(filters: List[Filter]) extends ForcedFilter with CombiningFilter {
+  override def operator: FilterOperation = AndFilterOperation
+  override def field: String = "and"
+  override def isEmpty : Boolean = filters.isEmpty
+  val asValues: String = filters.map(_.asValues).mkString("(",") AND (",")")
+}
+
+case class PreRenderedAndFilter(filters: Iterable[String]) extends CombiningFilter {
```
patelh: Maybe just call it RenderedAndFilter, no need to prefix it with Pre ;)

patelh:
lgtm

PR #423, opened 2019-03-25
Add filterOperation helpers in Request Model, first draft

ryankwagner:
PR to add helper functions (Stage 2 on issue #392 ) in preparation for OrFilter support.

Currently passes All testing, and fixes previously undiscovered bugs (IE FieldEquality validation across incompatible Dimension tables)


core/src/test/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorTest.scala
```diff
@@ -140,7 +140,7 @@ CAST(ssfu0.campaign_id AS VARCHAR) = CAST(c1.c1_id AS VARCHAR)
       SELECT CAST(COALESCE(account_id, 0) as VARCHAR) advertiser_id, CAST(COALESCE(impressions, 0) as VARCHAR) mang_impressions
         FROM(SELECT account_id, SUM(impressions) impressions
           FROM s_stats_fact_underlying
-          WHERE (ad_group_id = account_id) AND (account_id = 12345) AND (stats_date >= '$fromDate' AND stats_date <= '$toDate')
```
patelh: why was account_id filter removed?  Doesn't look right.

patelh:
@ryankwagner Remember when making changes to requestmodel, the order of operations matters.  Keep the ordering the same as you move the logic out to methods as necessary.  Try not to create too many methods, it will be come difficult to determine the order of execution when looking at the code.

core/src/test/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorTest.scala
ryankwagner: In this case, account_id is already being called for FieldEquality, and then the second instance can't be added to the set of filters.  I can modify FieldEqualityFilter to ignore the passed in fields in terms of adding them to the list of reserved filters, if that would help.  This was picked up in rendering and validation, but I can limit it to validation.

ryankwagner: In this case, account_id = 12345 AND account_id == ad_group_id can be passed differently as both being == 12345.

In another case account_id > 1 and account_id == ad_group_id, keeping both makes more sense.

patelh:
It would be best to keep refactoring and code fixes in separate PRs so you can easily code review and verify the changes.

patelh:
merge conflict on this PR

PR #433, opened 2019-04-04
Switch Or/And/Outer filters to set & add proper JSONR/JSONW for this.

ryankwagner:


patelh:
They were List to maintain ordering.  Not sure if you really want to change it to Set.

ryankwagner:
I will switch Set to LinkedHashSet to preserve this behavior.

pranavbhole:
> I will switch Set to LinkedHashSet to preserve this behavior.

Still it wont preserve the original ordering by user, you can see druid tests has changed the order of the filters.

ryankwagner:
It is now all Lists, and ordering will reflect passed in ordering for that reason.

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -1223,7 +1225,14 @@ object Filter extends Logging {
       case OuterFilter(filters) =>
         makeObj(
            ("operator" -> toJSON(filter.operator.toString))
-            :: ("outerFilters" -> toJSON(filters))
+            :: ("outerFilters" -> {
+             val values: mutable.ListBuffer[JValue] = mutable.ListBuffer.empty[JValue]
```
patelh: do you still need to do this if they are still List[_] ?  wouldn't toJSON(filters)) work ?

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -1368,6 +1377,21 @@ object Filter extends Logging {
     }
   }
 
+  implicit def filterSetJSONR: JSONR[List[Filter]] = new JSONR[List[Filter]] {
```
patelh: You shouldn't need this.  json4s can serialize and deserialize lists, so you just need to define a JSONR[Filter] which is already there right?

ryankwagner: json4s can serde ::, which is the super of List.  List 'works' by extension, but it isn't a stable call.

pranavbhole: If you check the line no 1386, we are already deserializing the List[Filter], why do we need JSONR for List[Filter]? 
```val fil = OrFilter.applyJSON(fieldExtended[List[Filter]]("filterExpressions"))(json)```

pranavbhole: this just to fix the errors in intellij in FilterOperation class

PR #436, opened 2019-04-16
Pull fields

ryankwagner:
Add OrFilterMeta fixes so that Druid Generator can properly run queries with Fact and Dimension filters.

Includes an edit to OrFilterMeta creating an enumeration to differentiate between, Metric, Fact, and Dimension columns being filtered.
Instead of returning an error, this logic now splits an OrFilter into constituent OrFilters in their proper locations (lookups, post aggregation, outer dim)


core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -493,6 +495,59 @@ object RequestModel extends Logging {
               val orFilterMap : Map[Boolean, Iterable[Filter]] = orFilter.filters.groupBy(f => publicFact.columnsByAliasMap.contains(f.field) && publicFact.columnsByAliasMap(f.field).isInstanceOf[PublicFactCol])
               require(orFilterMap.size == 1, s"Or filter cannot have combination of fact and dim filters, factFilters=${orFilterMap.get(true)} dimFilters=${orFilterMap.get(false)}")
               allOrFilterMeta += OrFilterMeta(orFilter, orFilterMap.head._1)
+
+              for(filter <- orFilter.filters) {
```
pranavbhole: can you reuse it ? put it into separate function

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -158,6 +158,7 @@ case class RequestModel(cube: String
                         , requestedFkAliasToPublicDimensionMap: Map[String, PublicDimension]
                         , orFilterMeta: Set[OrFilterMeta]
                         , dimensionRelations: DimensionRelations
+                        , useNewFeature: Boolean = false
```
pranavbhole: it is not used anywhere. I think instead of flag, we can use Map to Store feature flags. 
Create a trait called MahaFeature and you can use map of MahaFeature-> Boolean

pranavbhole:
can you run the generated druid query on the box and validate it?

core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
```diff
@@ -4887,7 +4887,8 @@ class RequestModelTest extends FunSuite with Matchers {
     }
   }
 
-  test("create model should fail when using or filters with fact and dim filter combination") {
+  //No longer valid, we now allow both & split on the backend.
```
patelh: In that case, do we have a test which tests allowing both?

patelh:
Can we get a better description of what this PR is attempting to accomplish?

core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
ryankwagner: The test is in DruidQueryGeneratorTest & generates both.

ryankwagner:
Description Updated.

core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
patelh: RequestModel is meant to work across data sources.  If you are changing behavior for creating model, it should be tested here first.  Query generation tests are specific to each engine.

pranavbhole:
verified that druid query is running with inner fact or filter and outer lookup dimension or filters.

pranavbhole:
@patelh @patelh can you please take look ?

ryankwagner:
Failure is the same JsonOutputFormat transient issue as we've been getting.

PR #443, opened 2019-04-23
Add LikeFilter as High Cardinality & use this logic for inner Druid limiting.

ryankwagner:


ryankwagner:
The failure is JsonOutputFormatTest (again)

PR #445, opened 2019-04-24
Aggregate derived dim col

pranavbhole:
Implementation of Issue: https://github.com/yahoo/maha/issues/438


pranavbhole:
@patelh @pavanab4u  Can you please take look at  this PR?

core/src/main/scala/com/yahoo/maha/core/query/hive/HiveQueryGeneratorCommon.scala
```diff
@@ -136,6 +136,10 @@ abstract class HiveQueryGeneratorCommon(partitionColumnRenderer:PartitionColumnR
           val name = column.name
           val nameOrAlias = column.alias.getOrElse(name)
           renderColumnWithAlias(fact, column, alias, Set.empty, false)
+          val isAggregatedDimCol = isAggregateDimCol(column)
+          if (isAggregatedDimCol) {
+            // do nothing in the group by
+          } else
```
patelh: this is else if?

core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
```diff
@@ -250,6 +250,12 @@ object HiveExpression {
     def asString : String = s"MIN(${s.asString})"
   }
 
+  case class COUNT(s: HiveExp, isDistinct: Boolean = false) extends BaseHiveExpression {
```
pavanab4u: better create another class as COUNT_DISTINCT rather than using boolean flag

core/src/main/scala/com/yahoo/maha/core/dimension/Dimension.scala
```diff
@@ -159,6 +165,37 @@ object HiveDerDimCol {
...
+                         dataType: DataType,
+                         columnContext: ColumnContext,
+                         derivedExpression: HiveDerivedExpression,
+                         alias: Option[String],
+                         annotations: Set[ColumnAnnotation],
+                         filterOperationOverrides: Set[FilterOperation]) extends BaseDerivedAggregateDimCol with WithHiveEngine {
+
+  require(derivedExpression != null,
+    s"Derived expression should be defined for a derived column $name")
+  require(derivedExpression.expression.hasRollupExpression, s"HiveDerDimAggregateCol should have rollup expression  $name - $derivedExpression")
```
patelh: You need to ensure the expression only references other dim cols, not fact cols.

pranavbhole: yup , very good point. I will add validation.

pranavbhole: added validation

core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
pranavbhole: Added COUNT_DISTINCT

core/src/main/scala/com/yahoo/maha/core/query/hive/HiveQueryGeneratorCommon.scala
pranavbhole: fixed

pranavbhole:
Addressed comments

PR #447, opened 2019-04-30
Add RowGrouping to fix Row Merging on MultiEngine Queries

ryankwagner:


patelh:
Dim driven or fact driven?

ryankwagner:
This is on queries that merge a Druid fact + Oracle Dim query.  Currently, RowList addRow uses only a single PrimaryKey alias for merging dim into fact.  This causes any multi-value Dimension Rows coming from the Fact (Pricing Type) to collapse into 1 row.  The fix is creating a RowGrouping object that contains both the fact-dim join key as its indexAlias, as well as any Fact-side Dimension Columns as "group by" keys to preserve the primary row list.

patelh:
You need to first define the use case which you are solving this for.  Multi-engine queries are there for particular use cases.  The behavior of multi-engine is dependent on fact driven vs dim driven queries so you need to be certain what you are solving.

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -44,9 +44,20 @@ object QueryPipeline extends Logging {
   val asyncDisqualifyingSet: Set[Engine] = Set(DruidEngine)
 
   val completeRowList: Query => RowList = (q) => new CompleteRowList(q)
-  val dimDrivenPartialRowList: Query => RowList = (q) => new DimDrivenPartialRowList(q.queryContext.indexAliasOption.get, q)
-  val dimDrivenFactOrderedPartialRowList: Query => RowList = (q) => new DimDrivenFactOrderedPartialRowList(q.queryContext.indexAliasOption.get, q)
-  val factDrivenPartialRowList: Query => RowList = (q) => new FactDrivenPartialRowList(q.queryContext.indexAliasOption.get, q)
+  val dimDrivenPartialRowList: Query => RowList = (q) => {
+    val indexAlias = q.queryContext.indexAliasOption.get
+    val groupByKeys: List[String] = q.queryContext.factGroupByKeys diff List(indexAlias)
+    new DimDrivenPartialRowList(RowGrouping(q.queryContext.indexAliasOption.get, groupByKeys), q)
```
patelh: This breaks the meaning of dim driven queries.  Dim driven queries will always have only 1 row for each key from driving dim.

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -44,9 +44,20 @@ object QueryPipeline extends Logging {
...
+  }
+  val dimDrivenFactOrderedPartialRowList: Query => RowList = (q) => {
+    val indexAlias = q.queryContext.indexAliasOption.get
+    val groupByKeys: List[String] = q.queryContext.factGroupByKeys diff List(indexAlias)
+    new DimDrivenFactOrderedPartialRowList(RowGrouping(q.queryContext.indexAliasOption.get, groupByKeys), q)}
```
patelh: Same here, dim driven implies single row per driving dim key.

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
ryankwagner: Thanks for the catch, that actually saved me some debugging.  :)

ryankwagner:
I have updated test cases to reflect the new logic.  Right now, the join cases for any query using multiple engines then joining the results is incorrect.  This adds columns to group by in order to prevent multivalue Dimension columns in fact from being lost.
The test cases in OracleQueryGenerator shows the fix, but current Query behavior is:
selectFields: ["Ad ID", "Ad Title", "Pricing Type"]
returnedRows from Fact: [1, null, "CPC"], [1, null "CPA"]
returnedRows from Dim: [1, "title", null]
Merged returned rows (current): [1, "title", "CPA"]
That is to say, the first of the two rows gets updated and CPC is lost.  The new logic creates a grouping on Pricing Type so that we get 2 returned rows instead of 1, both having the correct information.

core/src/test/scala/com/yahoo/maha/core/query/DefaultQueryPipelineFactoryTest.scala
```diff
@@ -501,17 +501,23 @@ class DefaultQueryPipelineFactoryTest extends FunSuite with Matchers with Before
         row.addValue("Ad Group ID", 202)
         row.addValue("Impressions", 100)
         row.addValue("Clicks", 1)
+        row.addValue("Pricing Type", "CPC")
```
patelh: Please create separate test for new functionality.  Should be easy to copy/paste the parts you need.

core/src/test/scala/com/yahoo/maha/core/query/DefaultQueryPipelineFactoryTest.scala
```diff
@@ -501,17 +501,23 @@ class DefaultQueryPipelineFactoryTest extends FunSuite with Matchers with Before
...
+        row2.addValue("Pricing Type", "CPA")
+        rl.addRow(row2)
     }.withOracleCallback {
       rl =>
         val row = rl.newRow
         row.addValue("Advertiser ID", 1)
         row.addValue("Ad Group Status", "ON")
         row.addValue("Campaign ID", 10)
         row.addValue("Ad Group ID", 202)
-        row.addValue("Source", 2)
```
patelh: This test is running a dim driven query.  No reason to modify this test.  It works as expected.

core/src/test/scala/com/yahoo/maha/core/query/DefaultQueryPipelineFactoryTest.scala
```diff
@@ -708,16 +714,15 @@ class DefaultQueryPipelineFactoryTest extends FunSuite with Matchers with Before
...
         row.addValue("Clicks", 1)
+        row.addValue("Pricing Type", "CPC")
         rl.addRow(row)
     }.withOracleCallback {
       rl =>
         val row = rl.newRow
         row.addValue("Advertiser ID", 1)
         row.addValue("Ad Group Status", "ON")
         row.addValue("Ad Group ID", 10)
-        row.addValue("Source", 2)
```
patelh: Again, this test is for dim driven query.  You should not be modifying this test.  It works as expected.

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -847,18 +857,19 @@ OuterGroupBy operation has to be applied only in the following cases
 
     def runMultiEngineQuery(factBestCandidateOption: Option[FactBestCandidate], bestDimCandidates: SortedSet[DimensionBundle], queryGenVersion: Version): QueryPipelineBuilder = {
       val indexAlias = bestDimCandidates.last.publicDim.primaryKeyByAlias
+      val factGroupByKeys = requestModel.bestCandidates.get.dimColMapping.values.toList
```
patelh: runMultiEngineQuery method is meant to run dim driven queries, not fact driven.

core/src/test/scala/com/yahoo/maha/core/query/DefaultQueryPipelineFactoryTest.scala
ryankwagner: This test is getting a Row with Pricing Type, Destination URL, and Source from a Dimension that does not contain any of those keys.  The returned Row List was invalid in the first place.

core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
ryankwagner: The group by keys are there for the Fact subqueries.  I will modify the rendered Dimension queries to ignore this value.

core/src/test/scala/com/yahoo/maha/core/query/DefaultQueryPipelineFactoryTest.scala
ryankwagner: More concretely,
`println(pipeline.queryChain.subsequentQueryList.head.asInstanceOf[OracleQuery].asString)`

Prints
`(SELECT * FROM (SELECT D.*, ROWNUM AS ROW_NUMBER FROM (SELECT  *
      FROM (SELECT ago0.advertiser_id "Advertiser ID", ago0."Ad Group Status" "Ad Group Status", ago0.id "Ad Group ID"
            FROM 
                (SELECT  DECODE(status, 'ON', 'ON', 'OFF') AS "Ad Group Status", advertiser_id, id
            FROM ad_group_oracle
            WHERE (advertiser_id = 213) AND (id IN (10))
             ) ago0 
           

           )
            ) D )) UNION ALL (SELECT * FROM (SELECT D.*, ROWNUM AS ROW_NUMBER FROM (SELECT * FROM (SELECT  *
      FROM (SELECT ago0.advertiser_id "Advertiser ID", ago0."Ad Group Status" "Ad Group Status", ago0.id "Ad Group ID"
            FROM 
                (SELECT  DECODE(status, 'ON', 'ON', 'OFF') AS "Ad Group Status", advertiser_id, id
            FROM ad_group_oracle
            WHERE (advertiser_id = 213) AND (id NOT IN (10))
             ) ago0 
           

           )
            ) WHERE ROWNUM <= 100) D ) WHERE ROW_NUMBER >= 1 AND ROW_NUMBER <= 100)`

patelh: The point is, it is hard to verify your changes and review them if you keep modifying tests that have nothing to do with new functionality.  As I noted in another comment, keep bug fixes or test fixes separate from new functionality.  What you are doing just makes it harder to review your code and makes it difficult to be confident about the changes not breaking other functionality.

core/src/test/scala/com/yahoo/maha/core/query/oracle/OracleQueryGeneratorTest.scala
```diff
@@ -5295,7 +5296,7 @@ class OracleQueryGeneratorTest extends BaseOracleQueryGeneratorTest {
          |            FROM
          |               ( (SELECT  advertiser_id, campaign_id, DECODE(status, 'ON', 'ON', 'OFF') AS "Ad Group Status", id
          |            FROM ad_group_oracle
-         |            WHERE (advertiser_id = 213) AND (id IN (45,39,30,51,48,27,33,54,12,15,42,36,21,18,24,53,41,35,17,50,44,23,38,47,26,11,32,14,20,29,46,52,28,34,55,40,49,43,22,16,37,19,25,31,13))
+         |            WHERE (advertiser_id = 213) AND (id IN (45,34,12,51,19,23,40,15,11,44,33,22,55,26,50,37,13,46,24,35,16,48,21,54,43,32,49,36,39,17,25,14,47,31,53,42,20,27,38,18,30,29,41,52,28))
```
ryankwagner: order changes since rows are now stored in a different order, as rowGrouping modifies mapping positions vs indexAlias key.

Issue #448, opened 2019-04-30
MultiEngineQuery Collapses Non-Pk Fact Table Dimension Rows


ryankwagner:
This issue is illustrated here:
https://github.com/yahoo/maha/pull/447

It is most noticeable when the column is Mapped (Pricing Type, for example), which makes query behavior inconsistent with the Reporting Request.

Given a Dimension located in the fact with multiple expected useful output values, this dimension gets collapsed as it is not the most granular key alias on the table.

patelh:
Is this for a dim driven query or fact driven?

PR #449, opened 2019-04-30
Add debug statement for multiEngine query profiling.

ryankwagner:


panditsurabhi:
Not required. Issue https://github.com/yahoo/maha/issues/448 is being worked on.

PR #453, opened 2019-05-07
Add coverage tests & fix tests using Fact columns in Dim rows.

ryankwagner:


core/src/test/scala/com/yahoo/maha/core/query/BaseQueryChainTest.scala
```diff
@@ -55,6 +56,37 @@ trait BaseQueryChainTest {
     qec
   }
```
ryankwagner: Below function should return no rows from Primary query so that Fallback Query code can be reached.

core/src/test/scala/com/yahoo/maha/core/query/MultiQueryTest.scala
```diff
@@ -24,5 +24,32 @@ class MultiQueryTest extends FunSuite with Matchers with BaseQueryGeneratorTest
...
+    val irlFn = (q : Query) => new DimDrivenPartialRowList("Advertiser ID", q)
+    val result = qc.execute(queryExecutorContext, irlFn, QueryAttributes.empty, new EngineQueryStats)
+    result.rowList.foreach {
+      r =>
+        model.requestCols.map(_.alias).foreach {
+          col => assert(r.getValue(col) === s"$col-value")
+        }
+    }
+  }
+
```
ryankwagner: Below executes query so that the logger logs a fallback query attempt (untestable code otherwise).

PR #467, opened 2019-06-13
Reduce lookups Leader logic

ryankwagner:
Implement initial Kafka Leader/Follower logic to be used for reduction of Oracle JDBC Lookups.

Noted in JdbcH2QueryTest, this adds functionality to notate a leader lookup node/nodes to perform JDBC lookups, then dump the results to a Kafka topic.

The next step is to set followers to read from this topic rather than from Oracle directly.

Finally, the Leaders no longer write data back to Druid & only write back to the Kafka topic.


patelh:
Not sure if it makes sense to change JDBC/Mongo lookups for this use case.  The point of JDBC/Mongo lookups is to read from those sources directly.  If you want to have one reader and write to Kafka, why not just use the existing RocksDB based lookups?  They already require updates be sent via Kafka.

ryankwagner:
In this case, the underlying goal is to reduce the number of JDBC connections being made on Oracle dimension tables for periodic Druid lookups.
Rather than having N nodes opening a JDBC connection to Oracle, some small subset of nodes should do the lookup & write back to Kafka and the rest should just read that topic.
This initial LOE doesn't demonstrate much more than use of Kafka, but reducing JDBC connections to Oracle is the end goal.

patelh:
@ryankwagner Yea, I understand what you are trying to do with having a leader for reading from JDBC and having other nodes read.  But this is exactly what the existing RocksDB based lookups do.  Please use existing functionality instead of modifying JDBC lookups to mimic similar behavior.  RocksDB already depends on protobuf, JDBC and Mongo do not and there is no reason to make them depend on protobuf, their schema is defined as config.  If you want to introduce new type of lookup, then go ahead, but as I already said, you can achieve the same with RocksDB lookups.  It is already being done for AdGroup and Ads, nothing stopping you from doing the same for other tables in Oracle.

patelh:
Adding a new type of lookup extraction namespace is a safer approach.  I still think you need to think through the leader/follower relationship and put it on paper so you can get feedback on design.

ryankwagner:
I have referenced the design for this on Issue 473

It elaborates on exact Leader logic as it is implemented in this PR, we well as how the next PR will look with Follower logic.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactoryWithLeaderAndFollower.java
```diff
@@ -0,0 +1,315 @@
...
+            };
+        }
+        if(extractionNamespace.getIsLeader()) {
+            return doLeaderOperations(id, extractionNamespace, lastVersion, cache, kafkaProperties, protobufSchemaFactory, lastDBUpdate);
+        } else {
+            return doFollowerOperations(id, extractionNamespace, lastVersion, cache, kafkaProperties, protobufSchemaFactory);
+        }
+    }
+
+    private List<Map<String, Object>> populateRowListFromJDBC(
```
pavanab4u: move this to superclass and use it here instead of duplicating.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactoryWithLeaderAndFollower.java
```diff
@@ -0,0 +1,341 @@
...
+    public Callable<String> doFollowerOperations(final String id,
+                                                 final JDBCExtractionNamespaceWithLeaderAndFollower extractionNamespace,
+                                                 final String lastVersion,
+                                                 final Map<String, List<String>> cache,
+                                                 final Properties kafkaProperties,
+                                                 final ProtobufSchemaFactory protobufSchemaFactory) {
+        LOG.info("Running Kafka Follower - Consumer actions on %s.", id);
+        String kafkaProducerTopic = extractionNamespace.getKafkaTopic();
+        kafkaConsumer = ensureKafkaConsumer(kafkaProperties);
+        kafkaConsumer.subscribe(Collections.singletonList(kafkaProducerTopic));
```
pavanab4u: consumer needs to be created just once

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/KafkaRowMapper.java
```diff
@@ -0,0 +1,56 @@
...
+        String rowPkCol = resultSet.getString(extractionNamespace.getPrimaryKeyColumn());
+
+        if(Objects.nonNull(cache))
+            cache.put(rowPkCol, strings);
+
+        Map<String, Object> row = new DefaultMapper().map(i, resultSet, statementContext);
+        if(Objects.nonNull(row)) {
+
+                ProducerRecord<String, byte[]> producerRecord =
+                        new ProducerRecord<>(producerKafkaTopic, extractionNamespace.getTable(), row.toString().getBytes());
```
pavanab4u: why aren't you using protobuf?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactoryWithLeaderAndFollower.java
```diff
@@ -0,0 +1,309 @@
...
+     * @param extractionNamespace
+     * @param cache
+     * @param value
+     */
+    public void updateLocalCache(final JDBCExtractionNamespace extractionNamespace, Map<String, List<String>> cache,
+                            final byte[] value) {
+
+        try {
+            String keyColumn = "";
+            String allColumnsString = new String(value).replaceAll("\\{", "").replaceAll("\\}", "");
```
pavanab4u: use protobuf

pavanab4u: what is this custom logic? just use hashmap serialize and deserialize

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactory.java
```diff
@@ -109,7 +90,33 @@ public Void withHandle(Handle handle) throws Exception {
         };
     }
 
-    private DBI ensureDBI(String id, JDBCExtractionNamespace namespace) {
+    Void populateRowListFromJDBC(
```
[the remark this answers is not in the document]
ryankwagner: As a callback, withHandle requires a Return even if it is null.
https://www.codota.com/code/java/methods/org.skife.jdbi.v2.DBI/withHandle

druid-lookups/src/test/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JdbcExtractionNamespaceCacheFactorySqlLocalQueryTest.java
```diff
@@ -0,0 +1,204 @@
...
+
+    @Mock
+    LookupService lookupService;
+
+    /**
+     * Set up local Kafka Producer and Comsumer, which will share properties.
+     */
+    void setKafkaProperties() {
+        kafkaProperties = new Properties();
+        kafkaProperties.put("bootstrap.servers", "localhost:9092");
```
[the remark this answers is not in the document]
ryankwagner: The local query test can use this since all internal tests are all disabled unless the user explicitly sets up Kafka & SQL to use them.  I will set these functions to only be called by their respective tests.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactoryWithLeaderAndFollower.java
```diff
@@ -0,0 +1,324 @@
...
+            final String id,
+            final JDBCExtractionNamespaceWithLeaderAndFollower extractionNamespace,
+            final String lastVersion,
+            final Map<String, List<String>> cache
+    ) {
+        kafkaProperties = extractionNamespace.getKafkaProperties();
+
+        Objects.requireNonNull(kafkaProperties, "Must first define kafkaProperties to create a JDBC -> Kafka link.");
+        final long lastCheck = lastVersion == null ? Long.MIN_VALUE / 2 : Long.parseLong(lastVersion);
+        if (!extractionNamespace.isCacheEnabled() && !extractionNamespace.getIsLeader()) {
```
pavanab4u: only condition should be checked here is !extractionNamespace.isCacheEnabled()

PR #475, opened 2019-07-10
add is image col to domain json

priyankadadlani:


priyankadadlani:
Creating another PR with other changes from forked branch.

PR #482, opened 2019-07-19
Double sum aggregator config

pranavbhole:


patelh:
It would be good to add some documentation for this flag and how to use our custom double sum agg.

pranavbhole:
yup will update documentation

PR #486, opened 2019-07-24
add other Engines which are part of MultiQueryEngineList only for SyncRequest

pavanab4u:


sapphirew:
👍

PR #515, opened 2019-09-03
Code to make Cache statements extensible in Lookups

ryankwagner:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/RocksDBExtractionNamespace.java
```diff
@@ -62,6 +67,16 @@ public RocksDBExtractionNamespace(@NotNull @JsonProperty(value = "namespace", re
         this.lookupAuditingEnabled = lookupAuditingEnabled;
         this.lookupName = lookupName;
         this.tsColumn = tsColumn;
+
+        try {
+            if(!cacheActionRunner.isEmpty() && Objects.nonNull(Class.forName(cacheActionRunner)))
```
panditsurabhi: Better check might be StringUtils.isNotBlank, since it checks for null too.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBExtractionNamespaceCacheFactory.java
```diff
@@ -69,70 +74,31 @@ public String call() {
...
 
+    public void tryResetRunnerOrLog(RocksDBExtractionNamespace extractionNamespace) {
+        try {
+            if (!extractionNamespace.cacheActionRunner.isEmpty()) {
+                cacheActionRunner = CacheActionRunner.class.cast(
+                        Class.forName(extractionNamespace.cacheActionRunner).newInstance());
+                LOG.error("Populated a new CacheActionRunner with description " + cacheActionRunner.toString());
+            }
+        } catch(Exception e){
+            LOG.error("Failed to get a valid cacheActionRunner.");
```
panditsurabhi: Please also log the error `"msg", e` to log the stacktrace

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBExtractionNamespaceCacheFactory.java
```diff
@@ -69,70 +74,31 @@ public String call() {
         };
     }
 
+    public void tryResetRunnerOrLog(RocksDBExtractionNamespace extractionNamespace) {
+        try {
+            if (!extractionNamespace.cacheActionRunner.isEmpty()) {
+                cacheActionRunner = CacheActionRunner.class.cast(
+                        Class.forName(extractionNamespace.cacheActionRunner).newInstance());
+                LOG.error("Populated a new CacheActionRunner with description " + cacheActionRunner.toString());
```
panditsurabhi: This is not an error log.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBExtractionNamespaceCacheFactory.java
```diff
@@ -69,70 +74,31 @@ public String call() {
...
-                Long newLastUpdated = Long.valueOf(newMessage.getField(field).toString());
-
-                final RocksDB db = rocksDBManager.getDB(extractionNamespace.getNamespace());
-                if (db != null) {
-                    byte[] cacheValue = db.get(key.getBytes());
-                    if(cacheValue != null) {
-
-                        Message messageInDB = parser.parseFrom(cacheValue);
-                        Long lastUpdatedInDB = Long.valueOf(messageInDB.getField(field).toString());
+        tryResetRunnerOrLog(extractionNamespace);
```
panditsurabhi: Everytime updateCache is called, the runner will be instantiated if it's defined. Can we add a check to do it only once in a running process?

pranavbhole:
can you elaborate the use case please?

ryankwagner:
We intend to have RocksDB-type lookup behavior to be controllable by configuration.  By default, the lookup would act as before, but the intent is that the rocksDB fetch and put behavior should be configurable by means of passing unique DB runners.  This would allow the user to create their own RocksDB caching implementation unique of what is currently in place.  That is, a user shouldn't be tied to our implementation of this lookup.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/RocksDBExtractionNamespace.java
```diff
@@ -62,6 +68,16 @@ public RocksDBExtractionNamespace(@NotNull @JsonProperty(value = "namespace", re
...
         this.tsColumn = tsColumn;
+
+        try {
+            if(StringUtils.isNotBlank(cacheActionRunner) && Objects.nonNull(Class.forName(cacheActionRunner)))
+                this.cacheActionRunner = cacheActionRunner;
+            else {
+                this.cacheActionRunner = "";
+            }
+        } catch (Exception e) {
+            this.cacheActionRunner = "";
```
panditsurabhi: Let's log this exception. An exception here would occur only if the given class is invalid or can't be found, which should not be ignored. In the current flow, it will throw a NullPointerException since tryResetRunnerOrLog method doesn't handle the case of empty string.

panditsurabhi: Actually let's just surface this exception, the application shouldn't start if the desired class isn't present.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBExtractionNamespaceCacheFactory.java
```diff
@@ -69,70 +75,33 @@ public String call() {
         };
     }
 
+    public void tryResetRunnerOrLog(RocksDBExtractionNamespace extractionNamespace) {
+        try {
+            if (!extractionNamespace.cacheActionRunner.isEmpty() && Objects.isNull(cacheActionRunner)) {
+                cacheActionRunner = CacheActionRunner.class.cast(
+                        Class.forName(extractionNamespace.cacheActionRunner).newInstance());
+                LOG.debug("Populated a new CacheActionRunner with description " + cacheActionRunner.toString());
+            } else if (Objects.isNull(cacheActionRunner)) {
```
panditsurabhi: We need an 'else' statement here, otherwise we risk NPE if neither condition is met.

PR #518, opened 2019-09-11
Druid lookups with Presto

sapphirew:
Enabling Presto lookup with following changes:
- Added `CustomizedTimestampMapper ` which can convert String/Long values to a Timestamp
- Added `TsColumnConfig.java` which includes additional settings for tsColumn
- Added `KerberosPreoperties` for JDBC connections using kerberos
- Made `JDBCExtractionNamespace.java` more generic to support JDBC connections besides oracle:
 - - Additional condition in `FIRST_TIME_CACHING_WHERE_CLAUSE ` and `SUBSEQUENT_CACHING_WHERE_CLAUSE` that selects only the latest partition of the table.
 - - Changed hardcoded `lastupdated` column name to tsColumn in the namespace class
- Removed duplicate fields of parent class in `JDBCExtractionNamespaceWithLeaderAndFollower `
 
A sample config json for changes in this PR:
```
{
	"version": "v0",
	"lookupExtractorFactory": {
		"type": "cachedNamespace",
		"extractionNamespace": {
			"type": "mahajdbc",
			"lookupName": "advertiser_lookup",
			"connectorConfig": {
				"createTables": false,
				"connectURI": "jdbc:presto://presto.path:4443/path/path"
			},
			"kerberosProperties": {
				"user": "user1",
				"SSL": "true",
				"SSLTrustStorePath": "/path/to/cert",
				"SSLTrustStorePassword": "changeit",
				"KerberosRemoteServiceName": "HTTP",
				"KerberosPrincipal": "user1@a.b.com",
				"KerberosUseCanonicalHostname": "false",
				"KerberosConfigPath": "/path/to/krb5.conf",
				"KerberosKeytabPath": "/home/user1/user1.keytab"
			},
			"table": "advertiser",
			"columnList": [
				"id",				
				"name"
			],
			"primaryKeyColumn": "id",
			"tsColumn": "last_updated",
			"tsColumnConfig": {
				"name": "last_updated",
				"type": "bigint",
				"format": "yyyyMMddhhmm",
				"secondaryTsColumn": "load_time",
				"secondaryTsColumnCondition": "="
			},

			"pollPeriod": "PT15M",
			"cacheEnabled": true
		},
		"firstCacheTimeout": 600000
	}
}
```


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CustomizedTimestampMapper.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    protected Timestamp extractByIndex(ResultSet r, int index) throws SQLException {
+        Timestamp result;
+
+        try {
+            result = r.getTimestamp(index);
+        } catch (Exception e) {
+            if (e.getMessage().contains(BIGINT_ERROR_MSG)) {
```
panditsurabhi: can we do r.getObject(index) and then check the type of object returned, instead of adding logic in exception block?

sapphirew: Good point.. will do

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CustomizedTimestampMapper.java
```diff
@@ -0,0 +1,59 @@
...
+
+    protected Timestamp getTimestampFromString(ResultSet r, int index) throws SQLException {
+        Timestamp result;
+        Date parsedDate;
+        String date = r.getString(index);
+
+        try {
+            parsedDate = dateFormat.parse(date);
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e.getMessage());
```
panditsurabhi: We're losing the stacktrace here. Do we need to catch and rethrow this exception?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactory.java
```diff
@@ -96,73 +89,146 @@ Void populateRowListFromJDBC(
...
+    }
+
+    private Object getTsValue(JDBCExtractionNamespace extractionNamespace, Timestamp updateTS) {
+        Object tsValue = updateTS;
+        if (extractionNamespace.hasTsColumnConfig()) {
+            if (extractionNamespace.getTsColumnConfig().isVarchar()) {
+                tsValue = new SimpleDateFormat(extractionNamespace.getTsColumnConfig().getFormat()).format(updateTS);
+            } else if (extractionNamespace.getTsColumnConfig().isBigint()) {
+                tsValue = updateTS.getTime();
+            }
```
panditsurabhi: Let's add an else here with a warning of unknown type

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CustomizedTimestampMapper.java
sapphirew: Yes.. `ParseException` must be caught here. I'll re-throw other exceptions.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactory.java
sapphirew: sure, will do

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactory.java
```diff
@@ -96,73 +89,146 @@ Void populateRowListFromJDBC(
...
+
         if (dbi == null) {
-            final DBI newDbi = new DBI(
-                    namespace.getConnectorConfig().getConnectURI(),
-                    namespace.getConnectorConfig().getUser(),
-                    namespace.getConnectorConfig().getPassword()
-            );
-            dbiCache.putIfAbsent(key, newDbi);
-            dbi = dbiCache.get(key);
+            if (!namespace.hasKerberosProperties()) {
```
panditsurabhi: hasKerberosProperties checks for size != 0, so even if the json is empty "{}", this logic will attempt to connect with kerberos. We can either have an explicit config to say which auth should be used (userid/password or kerberos) or enhance the check for 'hasKerberosProperties'.

sapphirew: if the json is empty "{}", the size of `KerberosProperties` would be 0.. However, having an explicit config sounds a better idea. Will add another field for that.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/JDBCExtractionNamespaceCacheFactory.java
sapphirew: throwing an exception instead if it's an unknown type

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CustomizedTimestampMapper.java
sapphirew: since I’m not able to modify the signature of extractByIndex in the parent class, I can only catch it and throw a new one

sapphirew: changed to handle `ParseException` only

PR #524, opened 2019-09-20
Removing prestoEngine for outer group by condition

pkdhiman:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
@@ -728,7 +728,14 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
                                     requestModel: RequestModel,
                                     queryAttributes: QueryAttributes,
                                     queryGenVersion: Version): Query = {
-    val queryType = if (isOuterGroupByQuery(bestDimCandidates, bestFactCandidate, requestModel)) {
+    val isOgb = isOuterGroupByQuery(bestDimCandidates, bestFactCandidate, requestModel)
+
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -728,7 +728,14 @@ class DefaultQueryPipelineFactory(implicit val queryGeneratorRegistry: QueryGene
                                     requestModel: RequestModel,
                                     queryAttributes: QueryAttributes,
                                     queryGenVersion: Version): Query = {
-    val queryType = if (isOuterGroupByQuery(bestDimCandidates, bestFactCandidate, requestModel)) {
+    val isOgb = isOuterGroupByQuery(bestDimCandidates, bestFactCandidate, requestModel)
```
panditsurabhi: Can we add a test with a request that is ogb, force engine Presto and make sure it falls back to Hive?

core/src/test/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorTest.scala
```diff
@@ -347,4 +347,32 @@ ORDER BY mang_impressions ASC
...
+      }"""
+
+    val request: ReportingRequest = getReportingRequestAsync(jsonString)
+    val registry = getDefaultRegistry()
+    val requestModel = RequestModel.from(request, registry)
+
+    assert(requestModel.isSuccess, requestModel.errorMessage("Building request model failed"))
+
+
+    val queryPipelineTry = generatePipelineForQgenVersion(registry, requestModel.toOption.get, Version.v2)
```
pranavbhole: you can remove line and generate query on against v0. Theres no V2 version of Presto generator

PR #553, opened 2019-10-24
Row Count Support in DruidQueryGenerator

mcjyang:


mcjyang:
@pranavbhole @pavanab4u can you review this PR?
Some changes:
1. adding new type DruidRowCountFactCol
2. Modifying DruidQueryGenerator for generating wrap-up groupby query for count
3. Correct Syntax in maha-lookup document.

service/src/main/scala/com/yahoo/maha/service/curators/Curator.scala
```diff
@@ -300,23 +322,91 @@ case class RowCountCurator(protected val requestModelValidator: CuratorRequestMo
...
+                val parRequestResult: ParRequestResult = mahaService.executeRequest(mahaRequestContext.registryName
+                  , totalRowsRequest, mahaRequestContext.bucketParams, mahaRequestLogBuilder)
+
+                val populateRowCount: ParRequest[RequestResult] = parRequestResult.prodRun.map(parRequestLabel, ParFunction.fromScala {
+                  requestResult =>
+                    var count = 0
+                    val resRowList = requestResult.queryPipelineResult.rowList
+                    if (!resRowList.isEmpty) {
+                      resRowList.foreach(row => {
+                        count = Option(row.getValue(0)).getOrElse("0").toString.toInt
```
pavanab4u: rather than doing this use getTotalRowCount method from RowList. In Druid query generator you can use QueryRowList.ROW_COUNT_ALIAS as your field name instead of "Row Count".

PR #557, opened 2019-10-28
Print out any long superset generation times as INFO logs.

ryankwagner:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
@@ -1759,6 +1759,7 @@ case class PublicFactTable private[fact](name: String
       .mapValues(_.map(tpl => tpl._2)
       .to[SortedSet])*/
 
+  private val startTime = System.currentTimeMillis()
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
```diff
@@ -1769,6 +1770,14 @@ case class PublicFactTable private[fact](name: String
...
       .mapValues(_.map(tpl => tpl._2)
       .to[SortedSet])
+  private val endTime = System.currentTimeMillis()
+
+  private val totalTimeSecs: Long = (endTime - startTime) / 1000
+  private val totalMinsTaken: String = String.valueOf(totalTimeSecs / 60)
+  private val totalSecsTaken: String = String.valueOf(totalTimeSecs % 60)
+
+  if(totalTimeSecs > 2)
+    logger.error(s"Total time for superset generation for $name v$revision: ${totalMinsTaken}m${totalSecsTaken}s ")
```
pranavbhole: can you also print the size of the map ?

PR #564, opened 2019-11-01
Fix static mapping for ogb queries

ritvikjaiswal:


core/src/main/scala/com/yahoo/maha/core/query/hive/HiveOuterGroupByQueryGenerator.scala
```diff
@@ -506,11 +502,28 @@ abstract case class HiveOuterGroupByQueryGenerator(partitionColumnRenderer:Parti
           case DecType(_, _, _, _, _, _) =>
             s"""ROUND(COALESCE($finalAlias, 0L), 10)"""
           case IntType(_,sm,_,_,_) =>
-            s"""COALESCE($finalAlias, 0L)"""
+            if (sm.isDefined && isOuterGroupBy) {
```
pranavbhole: can you pleas share this code? almost duplicated 2 times

PR #576, opened 2019-11-13
Add more debug logs

panditsurabhi:


panditsurabhi:
We're seeing cases where Presto JDBC resultset is returning the **expected number of rows but all columns values are null**. The output datasize on Presto coordinator for the query does not match the size of rows fetched through JDBC. This log will help debug that.

PR #579, opened 2019-11-16
Fix aliasColumnMap for ogb queries

ritvikjaiswal:


core/src/main/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorV1.scala
```diff
@@ -287,8 +287,7 @@ class PrestoQueryGeneratorV1(partitionColumnRenderer:PartitionColumnRenderer, ud
       case DimColumnInfo(alias) =>
         columnAliasToColMap += queryBuilderContext.getDimensionColNameForAlias(alias) -> queryBuilderContext.getDimensionColByAlias(alias)
       case ConstantColumnInfo(alias, value) =>
-      //TODO: fix this
-      //constantColumnsMap += renderColumnAlias(alias) -> value
+        columnAliasToColMap += renderColumnAlias(alias) -> new DimCol(alias, StrType(), new ColumnContext, None, Set.empty, Set.empty)
```
pranavbhole: fetch the column from fact instead of defining it again

PR #582, opened 2019-11-27
Discarding rows that greater than maxrows size

mcjyang:


mcjyang:
@pavanab4u @pranavbhole

PR #595, opened 2020-02-05
Off Heap Row List

pranavbhole:
Problem: Async Request Executors can generate large data and may trigger huge gc clean up. We are currently storing all rows in memory. 

Off heap row list : To overcome this, writing a OffHeapRowList which will store first N (inMemRowCountThreshold) rows in memory and after that it stores into rocksdb instance. User of the rowlist will be completely agonistic about where the data is stored, and all iterator functions on the rowlist works as expected. Added kill phase in the rowlist lifecycle which will take care of cleaning up rocksdb instance. 

To use this feature, you need to pass additional OffHeapRowListConfig in QueryPipelineContext while instantiating QueryPipelineFactory.

Please review and merge.  @patelh @pavanab4u @panditsurabhi 

Thank you 
 

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
@@ -10,7 +10,6 @@ import com.yahoo.maha.rocksdb.{RocksDBAccessor, RocksDBAccessorBuilder}
 import com.yahoo.maha.serde.LongSerDe
 import grizzled.slf4j.Logging
 import org.apache.commons.io.{FileUtils, FilenameUtils}
-import org.slf4j.LoggerFactory
 
 import scala.collection.mutable.ArrayBuffer
 import scala.util.Try
@@ -23,8 +22,6 @@ case class OffHeapRowListConfig(tempStoragePathOption:Option[String], inMemRowCo
 case class OffHeapRowList(query: Query, config: OffHeapRowListConfig) extends QueryRowList with AutoCloseable with Logging {
 
   private[this] val tempStoragePath = OffHeapRowList.getAndValidateTempStorage(config.tempStoragePathOption)
-
-  private val logger = LoggerFactory.getLogger(classOf[OffHeapRowList])
 
   private[this] var rocksDBAccessorOption: Option[RocksDBAccessor[Long, Row]] = None
 
@@ -128,7 +125,7 @@ case class OffHeapRowList(query: Query, config: OffHeapRowListConfig) extends Qu
     }
 
     if (!listAttempt.isSuccess) {
-      logger.warn("Failed to get total row count.\n" + listAttempt)
+      warn("Failed to get total row count.\n" + listAttempt)
     }
 
     total_count
--- a/core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
@@ -4,12 +4,14 @@ import java.io.File
 import java.util.UUID
 import java.util.concurrent.atomic.AtomicLong
 
+import com.twitter.chill.ScalaKryoInstantiator
 import com.yahoo.maha.core.query.QueryRowList.ROW_COUNT_ALIAS
 import com.yahoo.maha.core.request.Parameter.RequestId
 import com.yahoo.maha.rocksdb.{RocksDBAccessor, RocksDBAccessorBuilder}
-import com.yahoo.maha.serde.LongSerDe
+import com.yahoo.maha.serde.{LongSerDe, SerDe}
 import grizzled.slf4j.Logging
 import org.apache.commons.io.{FileUtils, FilenameUtils}
+import org.rocksdb.CompressionType
 
 import scala.collection.mutable.ArrayBuffer
 import scala.util.Try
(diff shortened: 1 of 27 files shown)


core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
```diff
@@ -0,0 +1,194 @@
...
+/*
+    Created by pranavbhole on 2/3/20
+*/
+case class OffHeapRowListConfig(tempStoragePathOption:Option[String], inMemRowCountThreshold: Int = 10000)
+
+case class OffHeapRowList(query: Query, config: OffHeapRowListConfig) extends QueryRowList with AutoCloseable with Logging {
+
+  private[this] val tempStoragePath = OffHeapRowList.getAndValidateTempStorage(config.tempStoragePathOption)
+
+  private[this] var rocksDBAccessorOption: Option[RocksDBAccessor[Long, Row]] = None
```
patelh: Storing a Row could get expensive because you're storing the alias map with each row.  In the InMem version, they all refer to 1 instance of the alias map so its not expensive, but serializing it would mean you store the copy of it with each row.  You may want to only store the Array buffer in the row.

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
```diff
@@ -0,0 +1,194 @@
...
+    val uuid = if(params.contains(RequestId)) {
+      params.get(RequestId)
+    } else {
+      UUID.randomUUID().toString
+    }
+
+    new RocksDBAccessorBuilder(s"off-heap-rowlist-${uuid}", Some(tempStoragePath))
+      .addKeySerDe(LongSerDe)
+      .addValSerDe(RowSerDe)
+      .toRocksDBAccessor
```
patelh: Set the cacheSize and writeBufferSize, default values are high.  Maybe 10MB and 1MB.

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  def isEmpty : Boolean = {
+    if (atomicLongRowCount.get() < 0) {
+      true
+    } else false
+  }
+
+  override protected def kill(): Unit = {
+    info(s"called killed on OffHeapRowList, Stats:-> inMemRowCountThreshold: ${config.inMemRowCountThreshold}, sizeOfInMemStore: ${sizeOfInMemStore()}, sizeOfRocksDB:${sizeOfRocksDB()}")
```
patelh: Probably want this in debug mode right, not on every query ?

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
```diff
@@ -0,0 +1,194 @@
...
+
+case class OffHeapRowList(query: Query, config: OffHeapRowListConfig) extends QueryRowList with AutoCloseable with Logging {
+
+  private[this] val tempStoragePath = OffHeapRowList.getAndValidateTempStorage(config.tempStoragePathOption)
+
+  private[this] var rocksDBAccessorOption: Option[RocksDBAccessor[Long, Row]] = None
+
+  private[this] val atomicLongRowCount = new AtomicLong(-1)
+
+  private[this] val rocksDBRowCount = new AtomicLong(0) // To be used for verification and monitoring
```
patelh: Does this need to be an atomiclong?  Are multiple threads going to be adding rows to this row list concurrently?

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
```diff
@@ -0,0 +1,194 @@
...
+*/
+case class OffHeapRowListConfig(tempStoragePathOption:Option[String], inMemRowCountThreshold: Int = 10000)
+
+case class OffHeapRowList(query: Query, config: OffHeapRowListConfig) extends QueryRowList with AutoCloseable with Logging {
+
+  private[this] val tempStoragePath = OffHeapRowList.getAndValidateTempStorage(config.tempStoragePathOption)
+
+  private[this] var rocksDBAccessorOption: Option[RocksDBAccessor[Long, Row]] = None
+
+  private[this] val atomicLongRowCount = new AtomicLong(-1)
```
patelh: Does this need to be atomiclong?  Are multiple threads going to be writing to the row list concurrently?

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
patelh: Also may want to change compression to just LZ4 instead of LZ4HC which is what we default to.  For query processing, LZ4HC might be too costly.  You can experiment and decide.

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
pranavbhole: Created RowValue case class which only stores ArrayBuffer

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
pranavbhole: fixed, no need of atomic

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
pranavbhole: fixed, no need of atomic

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
pranavbhole: printing stats only at debug or rocksdb init case.

core/src/main/scala/com/yahoo/maha/core/query/OffHeapRowList.scala
pranavbhole: fixed

pranavbhole: will do poc on LZ4 vs LZ4HC for huge row list.  Right now set to LZ4

pranavbhole:
thank you @patelh for your comments, it was really helpful

PR #596, opened 2020-02-10
Add ability to get source primitives for all expressions

ryankwagner:

Currently, for a given column the process for finding all source column primitives is not built in, and often unsupported (rollups).  This handles that by giving derived expressions syntax for source primitives, and adding a method to find primitives in a rollup's ColumnContext (expressed in test).

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
@@ -862,13 +862,14 @@ trait DerivedExpression[T] {
   def getPrimitiveCols(colNames: Set[String]): Set[String] = {
     val cols: Set[Column] = colNames.map(name => columnContext.getColumnByName(name)).filter(col => col.isDefined).map(col => col.get)
     cols.flatMap(col => {
-      if(col.isDerivedColumn) {
-        col.asInstanceOf[DerivedColumn].derivedExpression.sourcePrimitiveColumns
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
```diff
@@ -849,6 +849,17 @@ trait DerivedExpression[T] {
     columnRegex.findAllIn(expression.asString).map(_.substring(1).replace("}","")).toSet
   }
```
pranavbhole: Add comments about what is primitive/real cols

core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
```diff
@@ -849,6 +850,29 @@ trait DerivedExpression[T] {
...
+   * table, not derived in any way.
+   */
+  lazy val sourcePrimitiveColumns: Set[String] = {
+    getPrimitiveCols(sourceColumns)
+  }
+
+  def getPrimitiveCols(colNames: Set[String]): Set[String] = {
+    val cols: Set[Column] = colNames.map(name => columnContext.getColumnByName(name)).filter(col => col.isDefined).map(col => col.get)
+    cols.flatMap(col => {
+      if((!col.isDerivedColumn && !col.isInstanceOf[FactCol]) || (col.isInstanceOf[FactCol] && !col.asInstanceOf[FactCol].hasRollupWithEngineRequirement)) {
```
pranavbhole: can you simplify  this?

PR #599, opened 2020-02-12
Allow additive rollup and derived columns, via passed through discarding

ryankwagner:


pranavbhole:
exposing discard is meaning less, why would user care about what to discard, we just want a set of primitive cols for given Der Col

PR #610, opened 2020-03-11
Add revision-by-fk with registryTest

ryankwagner:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
@@ -1522,7 +1522,7 @@ case class FactBuilder private[fact](private val baseFact: Fact, private var tab
                    , revision: Int = 0
                    , dimRevision: Int = 0
                    , dimToRevisionMap: Map[String, Int] = Map.empty
-                   ) : PublicFactTable = {
+                   ) : PublicFact = {
     new PublicFactTable(name
       , baseFact
       , dimCols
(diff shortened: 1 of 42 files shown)


core/src/main/scala/com/yahoo/maha/core/registry/Registry.scala
```diff
@@ -21,23 +21,23 @@ import scala.collection.{SortedSet, mutable}
 class RegistryBuilder{
 
   private var publicDimMap: Map[(String, Int), PublicDimension] = Map()
-  private var publicFactMap: Map[(String, Int), PublicFact] = Map()
+  private var publicFactMap: Map[(String, Int), PublicFactTable] = Map()
   private var dimColToKeySetMap: Map[String, Set[String]] = Map()
   private var dimColToDimensionSetMap: Map[String, Set[(String, Int)]] = Map()
 
-  def register(fact: PublicFact): RegistryBuilder = {
```
pranavbhole: we should not be replacing Trait PublicFact With the Derived class PublicFactTable,


https://github.com/yahoo/maha/blob/c81d62dbdf1fa11ac355fc096bcacbaf24e16360/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala#L1660

core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
```diff
@@ -1556,6 +1560,7 @@ case class FactBuilder private[fact](private val baseFact: Fact, private var tab
       , revision
       , publicFact.dimRevision
       , Some(publicFact)
+      , publicFact.dimToRevisionMap ++ dimToRevisionOverrideMap
```
pranavbhole: Can you add test case to make that new values from dimToRevisionOverrideMap overwritten in the old map.

pranavbhole:
test looks good

PR #613, opened 2020-03-24
support nested group by queries for time bucketed hyper unique aggregations

ishan8:
Creates a HyperUnique rollup that generates nested group by queries in order to support aggregating a HyperUnique metric with cardinality calculated at a time based granularity and using that in a derived column post aggregation.


core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
```diff
@@ -1100,6 +1151,11 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
                               renderColumnWithAlias(fact, sourceCol, name, forPostAggregator = true)
                             }
                         }
+                      case DruidHyperUniqueTimeBucketedRollup(_) => {
```
patelh, line 1154: Do you need this here if you are just doing what the default case is doing?

core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
```diff
@@ -1037,6 +1076,18 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
...
+          getAggregatorFactory(
+            dataType,
+            SumRollup,
+            alias,
+            columnAlias)
+        case DruidHyperUniqueTimeBucketedRollup(fieldName) =>
+          val hyperUniqueOutputName = s"_hyperUniqueOutput_$fieldName"
+          postAggregatorList += new HyperUniqueFinalizingPostAggregator(columnAlias, hyperUniqueOutputName)
+          getHyperUniqueAggregatorFactory(dataType, hyperUniqueOutputName, fieldName)
+        case DruidHyperUniqueRollup(fieldName) if (forOuterQuery) =>
```
patelh, line 1089: Do you have an updated test to test this change?

ishan8: I included `{"field": "Unique Ad IDs"}`, which uses a `DruidHyperUniqueRollup`, in the test case below to cover this code path and ensure outer query aggregations for this rollup are correctly generated.

core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
ishan8: Removed it

core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
ishan8: I included `{"field": "Unique Ad IDs"}`, which uses a `DruidHyperUniqueRollup`, in the test case below to cover this code path and ensure outer query aggregations for this rollup are correctly generated.

hiralsp:
@ishan8 can you rebase of master

pranavbhole:
@ishan8  Can you please rebase this? lets get this in

PR #616, opened 2020-03-27
Add mapping into request model processing

ryankwagner:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
@@ -446,6 +446,7 @@ object RequestModel extends Logging {
               if(publicFact.dimToRevisionMap.isEmpty) Map.empty
               else {
                 requestedAliasList
+                    .filter(publicFact.dimToRevisionMap.keys.toSet.contains(_))
                   .filter(reqCol => {
                       val pdName = registry.getDimColIdentity(reqCol).get.publcDimName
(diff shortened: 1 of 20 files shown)


core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -446,6 +446,7 @@ object RequestModel extends Logging {
               if(publicFact.dimToRevisionMap.isEmpty) Map.empty
               else {
                 requestedAliasList
+                    .filter(publicFact.dimToRevisionMap.keys.toSet.contains(_))
```
pranavbhole: it means dimToRevisionMap is compulsory map to be filled in every cube ? 
Should not be the case

ryankwagner: This only filters input requested cols so that it only checks the ones requested in the map.

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -447,8 +447,10 @@ object RequestModel extends Logging {
               else {
                 requestedAliasList
                   .filter(reqCol => {
+                    if (registry.getDimColIdentity(reqCol).isDefined) {
                       val pdName = registry.getDimColIdentity(reqCol).get.publcDimName
                       registry.getDimColIdentity(reqCol).isDefined && publicFact.dimToRevisionMap.contains(pdName) && registry.dimMap.contains(pdName, publicFact.dimToRevisionMap(pdName))
```
panditsurabhi: The forst condition becomes redundant after the if statement

PR #620, opened 2020-04-10
Or filter with Fact and DimTypes for Druid

shek1608:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
@@ -729,10 +729,17 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
         }
       }
 
+      var dimTypePresent = false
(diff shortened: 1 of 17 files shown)


core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
```diff
@@ -729,10 +728,13 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
         }
       }
 
-      queryContext.requestModel.orFilterMeta.foreach{
-        meta =>
-          //val dimMeta = meta.orFilter.filters.filter()
-          if(!(meta.filterType == MetaType.FactType)) outerQueryDimFilterList += FilterDruid.renderOrDimFilters(meta.orFilter.filters, aliases.toMap, cols, Option.empty, true)
+      val orFilterTypes = queryContext.requestModel.orFilterMeta.map {
```
pavanab4u, line 731: why do you need this? 
can't you just do 
`queryContext.requestModel.orFilterMeta.filter(f => f.filterType == MetaType.DimType).foreach {`

PR #622, opened 2020-04-14
Update rollup & subset with onwardsDate requirements

ryankwagner:

This is a backwards incompatible change, as users will be expected to either make intermediate rollups non-queryable (NoopSchema) or define an earliest queryable date (availableOnwardsDate).  This is because new rollups are expected to not have all of the data that the parent table has, at least initially.


hiralsp:
@ryankwagner @panditsurabhi Why was this added?  Can you guys please add some rationale and why it is a breaking change?

PR #626, opened 2020-04-27
implement support for Druid period granularity queries

ishan8:

Added Druid period granularity support (described in https://druid.apache.org/docs/latest/querying/granularities.html) so that we can query time dimensions at arbitrary granularities.

For example, using DRUID_TIME_FORMAT_WITH_PERIOD_GRANULARITY("YYYY-MM-dd", "P1W")) would generate the following as part of the dimensions spec:
```
{
    "outputName": "Week",
    "extractionFn": {
        "timeZone": "UTC",
        "asMillis": false,
        "granularity": "WEEK",
        "type": "timeFormat",
        "format": "YYYY-MM-dd"
    },
    "type": "extraction",
    "dimension": "__time",
    "outputType": "STRING"
}
```

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/DerivedFunction.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/DerivedFunction.scala
@@ -259,6 +259,26 @@ object DruidDerivedFunction {
     val sourceDimColName = "__time"
   }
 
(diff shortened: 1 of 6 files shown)


pranavbhole:
can you rebase with master, i can merge this

PR #627, opened 2020-04-29
implement NotBetween filter operation for Druid

ishan8:

**Context**
We have a usecase where we want to check whether a metric is outside a certain range. We can't currently do this with `Or` filters because filters are stored in a map between the field name and the filter, so we can't have two filters on the same field. Adding this `NotBetween` filter as a workaround instead of more invasive changes to the query generator. 

Using this filter would generate the following having clause:
```
"type": "or",
"havingSpecs": [{
  "type": "greaterThan",
  "aggregation": "Uniques",
  "value": 1000
}, {
  "type": "lessThan",
  "aggregation": "Uniques",
  "value": 1
}]
```


patelh:
Generally, we like to keep this kind of functionality the same across all query generators.  It is not specific to one engine.

PR #630, opened 2020-05-05
upgrade to druid 0.17.1

sapphirew:
**Summary**
--

- Upgrade Druid version to 0.17.1
- Deprecated `SelectQuery` and introducing `ScanQuery` instead. Pagination feature is not supported anymore - https://druid.apache.org/docs/latest/querying/select-query.html 
@patelh just in case you are not aware of this change
- Added `NullHandling.initializeForTests()` in unit tests that import druid libs
- Updated maha-druid-lookups local setup documentation


pavanab4u:
@patelh , I believe you are using `select`. latest version has been modified which will not have pagination - https://druid.apache.org/docs/latest/querying/select-query.html

can you please review? 

we need this PR to me merged for upgrading the cluster.

patelh:
@pavanab4u we don't really have a choice right?  They got rid of select so what can we do?

pavanab4u:
that's correct, you'll need to migrate to `scan`

Issue #634, opened 2020-05-11
Druid filter on facts not included in the list of request columns


ishan8:
We have a Druid query use case where we want to filter out rows where an aggregated metric is less than a specified threshold, but don't want that metric's column included in the list of output columns.

This is possible with dimension columns but not with fact columns. Only fact columns in the list of request columns are added to the list of fact aggregators.
```
    val factCols = queryContext.factBestCandidate.factColMapping.toList.collect {
      case (nonFkCol, alias) if queryContext.factBestCandidate.requestCols(nonFkCol) =>
        (fact.columnsByNameMap(nonFkCol), alias)
    }
```
https://github.com/yahoo/maha/blob/master/core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala#L1143

Trying to understand why we need the `queryContext.factBestCandidate.requestCols(nonFkCol)` check. Would there be any issues if we removed it in order to support filtering on non output fact columns?

patelh:
Trying to understand the use case, e.g. you want to exclude rows with clicks < 5, but you are saying you don't want to include the click count in output, just for filtering rows right?  Best way to show what you are trying to do is create a PR with a broken test case.

ishan8:
Yes, as you said in the example you're describing, we'd just want to exclude rows with clicks < 5, but not display the click count.

A query like this would currently generate no aggregations on the clicks column because it's not included in `selectFields` list, so the `having` clause generated for the filter on the aggregated fact wouldn't work either.
```
{
  "cube": "k_stats",
  "selectFields": [
    {"field": "Advertiser ID"},
    {"field": "Country"},
    {"field": "Revenue"},
  ],
  "filterExpressions": [
    {"field": "Day", "operator": "=", "value": "$fromDate"},
    {"field": "Advertiser ID", "operator": "=", "value": "12345"},
    {"field": "Clicks", "operator": ">", "value": "5"}
  ],
  "paginationStartIndex":20,
  "rowsPerPage":100
}
```
I can create a PR with a broken test case if you think it would help.

patelh:
Feel create to create a PR with your test case and the change you proposed above.

PR #645, opened 2020-05-14
enable filtering on aggregated facts not included in output columns

ishan8:

Adding the ability to filter aggregated facts that aren't included in the list of output columns. e.g. in the test case added -- excluding rows with impressions < 5, but not including the impressions count column in the output.

More details here: https://github.com/yahoo/maha/issues/634


core/src/test/scala/com/yahoo/maha/core/query/druid/DruidQueryGeneratorTest.scala
```diff
@@ -3327,5 +3327,32 @@ class DruidQueryGeneratorTest extends BaseDruidQueryGeneratorTest {
...
+    val result = queryPipelineTry.toOption.get.queryChain.drivingQuery.asInstanceOf[DruidQuery[_]].asString
+    println(result)
+    val json = """\{"queryType":"groupBy","dataSource":\{"type":"table","name":"fact1"\},"intervals":\{"type":"intervals","intervals":\[".*"\]\},"virtualColumns":\[\],"filter":\{"type":"and","fields":\[\{"type":"selector","dimension":"statsDate","value":".*"\},\{"type":"selector","dimension":"advertiser_id","value":"12345"\}\]\},"granularity":\{"type":"all"\},"dimensions":\[\{"type":"default","dimension":"id","outputName":"Keyword ID","outputType":"STRING"\}\],"aggregations":\[\{"type":"longSum","name":"Clicks","fieldName":"clicks"\},\{"type":"longSum","name":"Impressions","fieldName":"impressions"\}\],"postAggregations":\[\],"having":\{"type":"and","havingSpecs":\[\{"type":"greaterThan","aggregation":"Impressions","value":5\}\]\},"limitSpec":\{"type":"default","columns":\[\],"limit":120\},"context":\{"applyLimitPushDown":"false","userId":"someUser","uncoveredIntervalsLimit":1,"groupByIsSingleThreaded":true,"timeout":5000,"queryId":"abc123"\},"descending":false\}"""
```
pranavbhole, line 3354: not sure why keyword value is missing from dimension list, can you please check? create another issue if this is existing bug.

ishan8: Seems like an existing issue because there's around 50 tests here that have `{"field": "Keyword Value"}` in `selectFields` but not in any of the generated queries.

pranavbhole: checked about the Keyword Value, theres no druid dim candidate for this col

PR #648, opened 2020-05-16
Dim Only Query: validate schema required filters

pranavbhole:

### DIFF
--- a/core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
+++ b/core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
@@ -135,7 +135,7 @@ class RequestModelTest extends FunSuite with Matchers {
           PubCol("ad_group_id", "Ad Group ID", InEquality),
           PubCol("ad_id", "Ad ID", InEquality),
           PubCol("campaign_id", "Campaign ID", InEquality),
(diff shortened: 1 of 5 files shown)


core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
```diff
@@ -5819,5 +5819,29 @@ class RequestModelTest extends FunSuite with Matchers {
     println(res.failed.get.getMessage)
     res.failed.get.getMessage should startWith (s"requirement failed: ERROR_CODE:10007 (Ad Format Name) can't be used with advertiser schema in publicFact cube")
   }
+
+  test ("Dim Only query Schema required filter validation") {
```
panditsurabhi, line 5823: Can we add a test to ensure it doesn't fail when the Advertiser ID filter is present?

pranavbhole: working on it,

pranavbhole: added one positive case, found that we have many positive cases all over test classes, shouldnt be problem.

pranavbhole:
https://github.com/yahoo/maha/issues/649

PR #655, opened 2020-05-20
Use postgres driver to escape literal

patelh:
Add a new literal mapper which uses postgres driver to escape literal.

Update a few plugins and qualify the use of Module in order to compile with JDK 11.


patelh:
@pranavbhole @pavanab4u please review.  ESAPI does not have a postgres codec and we currently use Oracle codec for everything so adding a new postgres literal mapper using the postgres driver.

pranavbhole:
need to rebase

PR #656, opened 2020-05-21
Flat buffer support for Druid-Lookups: WIP

pranavbhole:


patelh:
Maybe we should see if there is a better way to configure the schema factory.  Maybe we should create a new version here so we can cleanly move from old to new.

patelh:
as far as rocksdbmanager and kafkamanager are concerned, they only care about serializing and deserializing and then getting access to the data by field name and value IIRC.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunner.java
```diff
@@ -15,29 +12,39 @@
...
             , ServiceEmitter emitter
-            , RocksDBExtractionNamespace extractionNamespace){
+            , RocksDBExtractionNamespace extractionNamespace) {
+
         try {
             if (!extractionNamespace.isCacheEnabled() && valueColumn.isPresent()) {
                 return lookupService.lookup(new LookupService.LookupData(extractionNamespace, key, valueColumn.get(), decodeConfigOptional));
             }
 
+            if (routeToNextChain(schemaFactory)) {
```
panditsurabhi, line 42: Not sure if chain of responsibility is the right pattern here, since there's no 'chaining' of steps. 'schemaFactory' object can be type flatbuffer, protobuf or some other derived class of BaseSchemaFactory. Falling back to protobuf silently may not be the expected behavior always.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunnerFlatBuffer.java
```diff
@@ -0,0 +1,124 @@
...
+import com.yahoo.maha.maha_druid_lookups.server.lookup.namespace.schema.BaseSchemaFactory;
+import com.yahoo.maha.maha_druid_lookups.server.lookup.namespace.schema.flatbuffer.FlatBufferSchemaFactory;
+import com.yahoo.maha.maha_druid_lookups.server.lookup.namespace.schema.flatbuffer.FlatBufferWrapper;
+import org.rocksdb.RocksDB;
+
+import java.util.Optional;
+
+public class CacheActionRunnerFlatBuffer implements BaseCacheActionRunnerChain {
+
+    private static final Logger LOG = new Logger(CacheActionRunner.class);
```
panditsurabhi, line 22: Let the logger be CacheActionRunnerFlatBuffer.class

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/RocksDBLookupExtractor.java
```diff
@@ -47,21 +45,21 @@
...
+    private BaseSchemaFactory schemaFactory;
     private KafkaManager kafkaManager;
     private ServiceEmitter serviceEmitter;
     private Cache<String, byte[]> missingLookupCache;
     private final byte[] extractionNamespaceAsByteArray;
 
     public RocksDBLookupExtractor(RocksDBExtractionNamespace extractionNamespace, Map<String, U> map,
                                   LookupService lookupService, RocksDBManager rocksDBManager, KafkaManager kafkaManager,
-                                  ProtobufSchemaFactory protobufSchemaFactory, ServiceEmitter serviceEmitter) {
+                                  BaseSchemaFactory schemaFactory, ServiceEmitter serviceEmitter) {
```
pavanab4u, line 56: how is this going to be instantiated? during query time this factory is used for deserialization. It is instantiated here https://github.com/yahoo/maha/blob/master/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/cache/MahaNamespaceExtractionCacheManager.java#L391
where concrete class is injected https://github.com/yahoo/maha/blob/master/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/cache/MahaNamespaceExtractionCacheManager.java#L73

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunner.java
pavanab4u: I second that. there is no chain of responsibility here. you should call appropriate CacheActionRunner based on the configuration in the ExtractionNamespace which is already done here https://github.com/yahoo/maha/blob/master/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/RocksDBExtractionNamespace.java#L80-L90

patelh:
You should be able to specify the protobuf message class or flatbuffer message class in the json for extraction namespace.  Best if you create a new RocksDBV2ExtractionNamespace and clean up.

patelh:
Right now we have to provide config for protobuf schema factory in server side config.  It makes it more cumbersome to use.  Do all configuration as part of the json.  If protobuf or flatbuffer message class doesn't exist, error out.  This way all the user has to do is drop the message format jar in the libs folder, add the maha lookup extension and config and do rest via lookup registration.

pranavbhole:
Sorry for not addressing comments since long, had to wait for Druid Upgrade.  Now resuming the work on this, I have just rebased and added the parallel namespace for FlatBuffer extraction with optional  MahaNamespaceExtractionConfig config.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/schema/flatbuffer/FlatBufferWrapper.java
```diff
@@ -0,0 +1,26 @@
...
+    public abstract Table getFlatBuffer(byte[] flatBufferBytes);
+
+    public String readFieldValue(String fieldName, byte[] flatBufferBytes) {
+        return readFieldValue(fieldName, getFlatBuffer(flatBufferBytes));
+    }
+
+    public abstract FlatBufferBuilder createFlatBuffer(Map<String, FlatBufferValue> nameToValueMap);
+
+    public byte[] toByteArr(ByteBuffer byteBuffer) {
+        byte[] arr = new byte[byteBuffer.remaining()];
```
panditsurabhi, line 21: Let's add a null check for byteBuffer

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunnerFlatBuffer.java
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                    if (newLastUpdated > extractionNamespace.getLastUpdatedTime()) {
+                        extractionNamespace.setLastUpdatedTime(newLastUpdated);
+                    }
+                    serviceEmitter.emit(ServiceMetricEvent.builder().build(MonitoringConstants.MAHA_LOOKUP_UPDATE_CACHE_SUCCESS, 1));
+                }
+            } catch (Exception e) {
+                LOG.error(e, "Caught exception while updating cache");
+                serviceEmitter.emit(ServiceMetricEvent.builder().build(MonitoringConstants.MAHA_LOOKUP_UPDATE_CACHE_FAILURE, 1));
+                e.printStackTrace();
```
panditsurabhi, line 104: Do we need printStackTrace since e is logged above?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/JDBCExtractionNamespace.java
```diff
@@ -123,6 +123,11 @@ public boolean isCacheEnabled() {
         return cacheEnabled;
     }
 
+    @Override
+    public boolean isFlatBufferNamespace() {
```
panditsurabhi, line 127: Instead of adding this method in every namespace, either:
1. Add default implementation in ExtractionNamespace interface with false
2. Better - create a method 'getNamespaceType' in ExtractionNamespace which returns appropriately for each namespace ("jdbc", "flatbuffer" etc)

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/cache/MahaNamespaceExtractionCacheManager.java
```diff
@@ -387,6 +390,8 @@ private LookupExtractor getLookupExtractor(final ExtractionNamespace extractionN
             return new JDBCLookupExtractorWithLeaderAndFollower((JDBCExtractionNamespaceWithLeaderAndFollower) extractionNamespace, map, lookupService);
         } else if (extractionNamespace instanceof JDBCExtractionNamespace) {
             return new JDBCLookupExtractor((JDBCExtractionNamespace) extractionNamespace, map, lookupService);
+        } else if (extractionNamespace instanceof RocksDBExtractionNamespace && extractionNamespace.getType() == ExtractionNameSpaceSchemaType.FlatBuffer) {
```
pranavbhole, line 393: @pavanab4u  this is the how it instantiate the RocksDBLookupExtractor supporting flatbuffer schema.

pranavbhole: I have added ExtractionNameSpaceSchemaType to all existing name spaces in order to state which schema it supports.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunner.java
pranavbhole: removed the chain of responsibility.

pranavbhole: and added new cache action runner.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunnerFlatBuffer.java
pranavbhole: fixed

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/RocksDBLookupExtractor.java
pranavbhole: fixed,  added the comment in the class MahaNamespaceExtractionCacheManager

pranavbhole:
> You should be able to specify the protobuf message class or flatbuffer message class in the json for extraction namespace. Best if you create a new RocksDBV2ExtractionNamespace and clean up.

Added the config in the RocksDBExtractionNamespace for accepting flat buffer cache runner. And added an optional config into MahaNamespaceExtractionConfig to configure the name of the flat buffer schema factory class  which is dumped into jar.

pranavbhole:
There are not breaking changes along with this PR, all configurations are optional.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/JDBCExtractionNamespace.java
```diff
@@ -123,6 +122,11 @@ public boolean isCacheEnabled() {
         return cacheEnabled;
     }
 
+    @Override
+    public ExtractionNameSpaceSchemaType getType() {
+        return ExtractionNameSpaceSchemaType.Protobuf;
```
pavanab4u, line 127: technically JDBCExtractionNamespace is not using any schema. it is storing String as it is. you may want to have `None` as additional type for this case.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/MongoExtractionNamespace.java
```diff
@@ -118,6 +118,11 @@ public boolean isCacheEnabled() {
         return cacheEnabled;
     }
 
+    @Override
+    public ExtractionNameSpaceSchemaType getType() {
+        return ExtractionNameSpaceSchemaType.Protobuf;
```
pavanab4u, line 123: same here. you need `None` type

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/BaseCacheActionRunner.java
```diff
@@ -0,0 +1,36 @@
...
+    void updateCache(BaseSchemaFactory schemaFactory
+            , final String key
+            , final byte[] value
+            , RocksDB db
+            , ServiceEmitter serviceEmitter
+            , RocksDBExtractionNamespace extractionNamespace);
+
+    void validateSchemaFactory(BaseSchemaFactory schemaFactory);
+
+    boolean flatBufferSupport();
```
pavanab4u, line 35: i think this is redundant. implementation type should tell what kind of schema it supports. you can just check `instanceof`

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/CacheActionRunner.java
```diff
@@ -78,14 +78,16 @@ public String handleDecode(DecodeConfig decodeConfig, Message parsedMessage, Des
...
-    synchronized public void updateCache(ProtobufSchemaFactory protobufSchemaFactory
+    synchronized public void updateCache(BaseSchemaFactory schemaFactory
             , final String key
             , final byte[] value
             , RocksDB db
             , ServiceEmitter serviceEmitter
             , RocksDBExtractionNamespace extractionNamespace) {
         if (extractionNamespace.isCacheEnabled()) {
             try {
+                validateSchemaFactory(schemaFactory);
```
pavanab4u, line 89: it would be better to use bounded type parameter and restrict this class to be used only with `ProtobufSchemaFactory`. with this it will be a compile time check as against current runtime check.

pranavbhole: schema factory is compile time validation now, had to add extra logic at the rocksdb namespace

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/entity/BaseCacheActionRunner.java
pranavbhole: removed this

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/MongoExtractionNamespace.java
pranavbhole: done

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/JDBCExtractionNamespace.java
pranavbhole: done

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/namespace/ExtractionNamespace.java
```diff
@@ -29,4 +28,6 @@
     String getTsColumn();
 
     boolean isCacheEnabled();
+
+    ExtractionNameSpaceSchemaType getType();
```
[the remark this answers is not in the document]
pranavbhole: done

PR #664, opened 2020-06-03
download and use new rocksDB instance if kafka is enabled for the namespace

mcjyang:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/KafkaManager.java
```diff
@@ -206,7 +206,11 @@ public Boolean call() {
                         if(seekOffsetToPreviousSnapshot) {
                             kafkaPartitionOffset.entrySet().forEach(partitionOffset -> {
                                 log.info("topic = [%s], seek partition = [%s], seek offset = [%s]", topic, partitionOffset.getKey(), partitionOffset.getValue());
-                                consumer.poll(0);
+                                try {
+                                    consumer.poll(0);
+                                } catch (Exception e) {
+                                    log.error(e, "Caught exception while trying to poll kafka event(s)");
+                                }
                                 consumer.seek(new TopicPartition(topic, partitionOffset.getKey()), partitionOffset.getValue());
```
pavanab4u, line 214: seek() is the what will throw error for invalid offsets. include this line also in your try block. also, in catch block set consumer to seekToBeginning

PR #665, opened 2020-06-04
Add date time between filter

patelh:
Add a date time between filter.  Currently we have a confusing array of Day, Hour, Minute filters which make it hard to understand how to query the data at hour or minute grain.  Adding a first class filter for date time makes this more straightforward.  In order to keep it simple, we only add a between filter since that's the most used and can also emulate an equality filter.

We move the time zone method out of UTC provider since it should be provided by a concrete user time zone provider.  It's not the UTC provider's function to provide the user time zone.  This is a breaking change but should be easy to migrate since the concrete implementation can just extend both UTCTimeProvider and UserTimeZoneProvider.

Added tests for all generators.  Note, I purposely changed the RequestModel.from function to require a UserTimeZoneProvider to ensure downstream changes are forced.

I was thinking of changing major versions since there are breaking changes: 6.1

### DIFF
--- a/api-example/pom.xml
+++ b/api-example/pom.xml
@@ -11,7 +11,7 @@
     <parent>
         <artifactId>maha-parent</artifactId>
         <groupId>com.yahoo.maha</groupId>
-        <version>5.383-SNAPSHOT</version>
+        <version>6.1-SNAPSHOT</version>
     </parent>
 
     <name>maha api-example</name>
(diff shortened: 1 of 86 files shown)


pranavbhole:
need to rebase

patelh:
@panditsurabhi @pranavbhole @pavanab4u This is ready for review.

patelh:
@panditsurabhi @pranavbhole @pavanab4u Don't mean to rush you, but it would be great if I can get some feedback by end of week so I can address it over weekend so we are good to get this merged on Monday/Tuesday next week if possible.

PR #669, opened 2020-06-19
use same update logic like listener while applyChangesSinceBeginning

mcjyang:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/KafkaManager.java
```diff
@@ -142,9 +138,13 @@ public Boolean call() {
                                 continue;
                             }
                             try {
-                                updateRocksDB(parser, descriptor, tsField, rocksDB, key, message);
```
pavanab4u, line 145: since `updateRocksDB` method is not going to be used you can delete this private method from the class.

PR #672, opened 2020-06-23
Add optional filter map by schema

ryankwagner:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
@@ -572,12 +572,12 @@ object RequestModel extends Logging {
               require(filterMap.contains(alias), s"Missing required filter: cube=${publicFact.name}, field=$alias")
           }
 
-          val optionalFilterColumns = publicFact.optionalFilterColumns
+          val requiredFilterColumns = publicFact.requiredFilterColumns
           require(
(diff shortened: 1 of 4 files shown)


core/src/test/scala/com/yahoo/maha/core/RequestModelTest.scala
```diff
@@ -5930,5 +5930,29 @@ class RequestModelTest extends FunSuite with Matchers {
     val res = RequestModel.from(request, registry)
     assert(res.isSuccess, s"should not fail on having filter on Advertiser ID")
   }
+
+  test("Should fail without any part of required filtering list") {
+    val jsonString = s"""{
```
panditsurabhi, line 5935: Can we add a test to make sure only one of the filter columns is required, not all.

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -572,6 +572,14 @@ object RequestModel extends Logging {
               require(filterMap.contains(alias), s"Missing required filter: cube=${publicFact.name}, field=$alias")
           }
 
+          val optionalFilterColumns = publicFact.optionalFilterColumns
```
panditsurabhi, line 575: Can we call this 'requiredFilterColumns' instead of 'optionalFilterColumns' and add a code comment in publicFact explaining that only one of them is required.

PR #676, opened 2020-06-30
add rocksDBNamespace column & modify lookupService behavior if overrideHostList present

mcjyang:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/LookupService.java
```diff
@@ -211,16 +215,19 @@ public void update(final LookupData lookupData, byte[] value) {
         String valueColumn;
         Optional<DecodeConfig> decodeConfigOptional = Optional.empty();
         ExtractionNamespace extractionNamespace;
+        List<String> overrideHostList;
```
pavanab4u, line 218: why do you want to expose this? extractionNamespace already has getOverrideLookupServiceHostsList() method

PR #687, opened 2020-08-06
Making Requestlog write async

pavanab4u:


request-log/src/main/scala/com/yahoo/maha/log/KafkaMahaRequestLogWriter.scala
```diff
@@ -90,7 +90,10 @@ class KafkaMahaRequestLogWriter(kafkaRequestLoggingConfig: KafkaRequestLoggingCo
   }
 
   def write(proto: MahaRequestProto) : Unit = {
-    writeMahaRequestProto(proto)
+    implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global
```
patelh, line 93: This feels hacky, why not create another method which takes in a executoncontext, e.g.
```
def writeAsync(proto: MahaRequestProto, ec: ExecutionContext): Unit
```
This way you can control the execution context and what to do when the pool is full.

patelh: Also, I don't see why it isn't async already.  The writeMahaRequestProto method already returns a future.  It's already async.

pranavbhole: it is actually problem of the kafka producer library, when kafka server is not available, it blocks the call wait for the response from server for "max block ms" this blocks the response in api. Thus wrapping it up in the another thread.

patelh: Yea, but the blocking behavior only occurs if the local queue is full.  All you are doing by using another thread pool is passing the blocking behavior into another thread pool.  The call in the other thread pool will block and soon that pool be exhausted and you'll have a bunch of submitted tasks in its queue.  Effectively causing a memory leak.  The max size of forkjoin pool is 64 million.

patelh: Ideally configure kafka client to timeout earlier or size the queue such that once queue is full it starts dropping messages.

pranavbhole: fixing

PR #690, opened 2020-08-07
Poweset storage optimization for Public fact

pranavbhole:
Problem: We generate the power for foreign keys in fact and store it in memory. If foreign keys exceeds count of 20, poweset size becomes millions (2 to the power of 20).  Hard to store it in memory. 

Solution: Defining the PowesetStorage trait for public facts, default one is Inmemory storage. RocksdbPoweset store push the entries out of memory.  
PowerSet is now storing SortedSet of Fks => SortedSet Fact Names. (instead of entire Fact). This makes it easy to SerDe in rocksdb. 

Please review: 
 @patelh @pavanab4u @panditsurabhi @shek1608

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
@@ -1848,20 +1848,7 @@ case class PublicFactTable private[fact](name: String
     if (this.parentFactTable.isDefined) parentFactTable.get.getSecondaryDimFactMap
     else
 */
-   if (getPowerSetStorage.isEmpty) {
-     facts
-      .values
-      .map(f => (f.dimCols.filter(_.annotations.exists(_.isInstanceOf[ForeignKey])).map(col => col.name), f.name))
-      .par
-      .flatMap(tpl => utils.power(tpl._1, tpl._1.size).map(s => (s.to[SortedSet], tpl._2)))
-      .toIndexedSeq
-      .groupBy(_._1)
-      .mapValues(_.map(tpl => tpl._2)
-      .to[SortedSet]).foreach {
-      entry=>
-        powerSetStorage.store(FactSearchKey(entry._1), entry._2)
-    }
-   }
+  getPowerSetStorage.store(facts.values)
 
   private[this] val dimColsByName = dimCols.map(_.name)
 
@@ -1953,7 +1940,7 @@ case class PublicFactTable private[fact](name: String
 
         val factsToSearch = {
           if(fkCols.nonEmpty) {
-            getPowerSetStorage.search(FactSearchKey(fkCols)).map(nameSet=> nameSet.map(name=> facts(name)))
+            getPowerSetStorage.search(fkCols)
           } else {
             Option(facts.values)
           }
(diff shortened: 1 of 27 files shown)


core/src/main/scala/com/yahoo/maha/core/fact/Fact.scala
```diff
@@ -1838,22 +1843,32 @@ case class PublicFactTable private[fact](name: String
...
       .values
-      .map(f => (f.dimCols.filter(_.annotations.exists(_.isInstanceOf[ForeignKey])).map(col => col.name), f))
+      .map(f => (f.dimCols.filter(_.annotations.exists(_.isInstanceOf[ForeignKey])).map(col => col.name), f.name))
       .par
       .flatMap(tpl => utils.power(tpl._1, tpl._1.size).map(s => (s.to[SortedSet], tpl._2)))
       .toIndexedSeq
       .groupBy(_._1)
       .mapValues(_.map(tpl => tpl._2)
-      .to[SortedSet])
+      .to[SortedSet]).foreach {
```
patelh, line 1860: This may force a materialization, try using a view and do all the transformations, and if you are storing underneath then why even sort it ?

core/src/main/scala/com/yahoo/maha/core/fact/PowerSetStorage.scala
```diff
@@ -0,0 +1,72 @@
...
+
+trait PowerSetStorage {
+  def store(searchKey: FactSearchKey, facts: SortedSet[String]): Unit
+  def search(searchKey: FactSearchKey) : Option[SortedSet[String]]
+  def size: Long
+  def isEmpty: Boolean
+}
+
+class DefaultPowerSetStorage extends PowerSetStorage {
+  private[this] val secondaryDimFactMap = new mutable.HashMap[SortedSet[String], SortedSet[String]]()
```
patelh, line 21: Is this HashMap thread safe?

core/src/main/scala/com/yahoo/maha/core/fact/PowerSetStorage.scala
```diff
@@ -0,0 +1,72 @@
...
+    RocksDBAccessor.cleanupBaseDir(baseDirOption.get)
+  }
+
+  private[this] val rocksDB = new RocksDBAccessorBuilder(s"PowerSetStorage_", baseDirOption)
+    .addKeySerDe(FactSearchKeySerDe)
+    .addValSerDe(FactSerchResultSerDe).toRocksDBAccessor // Using default settings for now
+  private[this] var count: Long = 0;
+
+  def store(searchKey: FactSearchKey, facts: SortedSet[String]): Unit = {
+    rocksDB.put(searchKey, facts)
```
patelh, line 52: Is the put thread safe on rocksDB?

patelh:
@pranavbhole any other way to prune this set, you may not need to store all possibilities...

patelh:
@pranavbhole another option could be to do this lazily, calculate on first use and put in an LRU.

patelh:
@pranavbhole consider using a RoaringBitmap here, https://github.com/RoaringBitmap/RoaringBitmap.  For each fact, you have a bitmap which contains all the keys it contains.  You'd have to create a foreign key mapping such that each key has an integer value to add to the bitmap for each fact.  You'd then have to change the search to do a contains on each bitmap for each fact for all the fks.

patelh:
May consider renaming since it's essentially a facts by foreign key search.

patelh:
@pranavbhole you may get better performance with a BitSet since number of foreign keys is going to be relatively small, < 1000

pranavbhole:
Thank you @patelh  very much. I have gone through roaring bitmap implementations. It is really perfect fit here. Thank you again for suggesting roaring bit map and adding the code. 

I was testing about the other solution where we store 
map of fk-> List of supported fact. 
Search by request fk cols and take intersection of sets from map. This also works but it has more computation at the request flow.  Bitmap will definitely going to work better.

pranavbhole:
Renamed the base trait

PR #692, opened 2020-08-11
Remove rowCount when not asked for

ryankwagner:


hiralsp:
@ryankwagner does this also need to be fixed for other engines?

ryankwagner:
@hiralsp I've only profiled it for Oracle & am taking comments for that, then moving on to other engines.  ATM the only engine sticking out to 'need' the fix is Oracle, but I will make tests for all others and name them as such.

The underlying issue that this fixes is that given a query for only Dims and the desire for DIstinct values, the outer select statement uses the phrase ROWNUM as ROW_NUMBER (noted in OracleQueryGeneratorTest line 2413, in this PR) which makes the distinct statement meaningless.

patelh:
@ryankwagner I'm always afraid these kind of issues will get lost and we'll have different engines optimized differently.

ryankwagner:
Checked Hiva, Druid generators.  Hive doesn't mention Pagination, Druid implements pagination via post processing in the executor.  Postgres has a wrapper for it, but it appears this wrapper uses native rownum which is the same optimization Oracle is gaining here.

PR #700, opened 2020-08-19
Revert "Remove rowCount when not asked for"

ryankwagner:
Reverts yahoo/maha#692

Temporarily revert this, as non-paginated nested queries in Oracle with a SI > 1 will fail.  The fix will be to include rownum as row_number on an actual wrapper mid-select (coming later)


ryankwagner:
Please wait for UNION ALL fix & travis upgrade

PR #703, opened 2020-08-20
adding info log for rocksDB lookup

sapphirew:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java
```diff
@@ -21,6 +21,7 @@
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hdfs.DistributedFileSystem;
+import org.eclipse.jetty.util.log.Log;
```
hiralsp, line 24: Is that the right logger?

sapphirew: mistakenly added by auto import. removed it

PR #717, opened 2020-09-21
add random rockDB path as an option in configuration

sapphirew:

### DIFF
--- a/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java
+++ b/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java
@@ -129,8 +129,8 @@ public class RocksDBManager {
             FileUtils.forceMkdir(file);
         }
 
-        final String localZippedFileNameWithPath = String.format("%s/%s/%s/rocksdb_%s.zip",
-                localStorageDirectory, extractionNamespace.getNamespace(), loadTime, UUID.randomUUID().toString());
+        final String localZippedFileNameWithPath = String.format("%s/%s/%srocksdb_%s.zip",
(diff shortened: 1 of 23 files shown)


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java
```diff
@@ -129,8 +129,8 @@ public String createDB(final RocksDBExtractionNamespace extractionNamespace,
             FileUtils.forceMkdir(file);
         }
 
-        final String localZippedFileNameWithPath = String.format("%s/%s/%s/rocksdb_%s.zip",
-                localStorageDirectory, extractionNamespace.getNamespace(), loadTime, UUID.randomUUID().toString());
+        final String localZippedFileNameWithPath = String.format("%s/%s/%srocksdb_%s.zip",
+                localStorageDirectory, extractionNamespace.getNamespace(), loadTime, getLocalPathSuffix(extractionNamespace.isRandomLocalPathSuffixEnabled()));
```
pranavbhole, line 133: Did you notice that there is one less slash in change, zip file is stored in parent dir with prefix as loadTime? Looks like bug. Need to add slash after load time

pranavbhole (changes requested):

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java
sapphirew: I removed the slash intentionally as the path suffix could be empty. The specific logic is on this line https://github.com/yahoo/maha/blob/0d76a8b08112e26dfbd399afb2b3942761c33a7c/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java#L165

pavanab4u (approved):

pranavbhole (approved):

PR #729, opened 2020-10-27
Support query with both aliases

lgadde:


core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -181,6 +181,21 @@ class QueryBuilderContext {
...
 
+  def containsColByNameAndAlias(name: String) : Boolean = {
+    if(containsColByName(name)){
+      var aliasName : Option[String] = None
+      if(colAliasToFactColNameMap.contains(name) || colAliasToDimensionColNameMap.contains(name)){
+        aliasName = colAliasToFactColNameMap.get(name)
+      }else if (colAliasToDimensionColNameMap.contains(name)){
+        aliasName = colAliasToDimensionColNameMap.get(name)
+      }else{
+        true
```
pranavbhole, line 192: redundant true value

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -181,6 +181,21 @@ class QueryBuilderContext {
...
+      }
+      if(aliasName.isDefined && containsColByName(aliasName.get)) true
+    }
+    false
```
pranavbhole, line 196: this method will always return false

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -181,6 +181,21 @@ class QueryBuilderContext {
   
   def containsColByName(name: String) : Boolean = columnNames.contains(name)
 
+  def containsColByNameAndAlias(name: String) : Boolean = {
+    if(containsColByName(name)){
+      var aliasName : Option[String] = None
+      if(colAliasToFactColNameMap.contains(name) || colAliasToDimensionColNameMap.contains(name)){
```
pranavbhole, line 187: || colAliasToDimensionColNameMap.contains(name) is redundant here

pranavbhole:
please add tests for all generators

pranavbhole (changes requested):

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
pranavbhole: need to fix

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
pranavbhole: need to fix

pranavbhole (changes requested):

pranavbhole (changes requested):

pranavbhole (approved):

PR #749, opened 2020-12-19
implement support for BigQuery

ishan8:
We had a use-case internally at Twitter to route queries between Druid and BigQuery, so I added support for BigQuery and wanted to submit it back here. Would be great if you could help review it!

---


bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,174 @@
...
+    val om = new ObjectMapper(new YAMLFactory()) with ScalaObjectMapper
+    om.registerModule(DefaultScalaModule)
+    om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+    om
+  }
+
+  def buildBigqueryClient(): BigQuery = {
+    BigQueryOptions
+      .newBuilder()
+      .setCredentials(ServiceAccountCredentials.fromStream(new FileInputStream(gcpCredentialsFilePath)))
```
patelh, line 54: Is the stream being closed here?

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,174 @@
...
+      .setProjectId(gcpProjectId)
+      .setRetrySettings(ServiceOptions.getDefaultRetrySettings().toBuilder().setMaxAttempts(retries).build())
+      .build()
+      .getService
+  }
+
+  def configureProxy(): Unit = {
+    if (enableProxy && proxyCredentialsFilePath.isDefined && proxyHost.isDefined && proxyHost.isDefined) {
+      val proxyCredentials = mapper.readValue[ProxyCredentials](
+        new FileInputStream(proxyCredentialsFilePath.get)
```
patelh, line 64: same here, is this being closed?

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,174 @@
...
+      )
+
+      System.setProperty(HttpsProxyHost, proxyHost.get)
+      System.setProperty(HttpsProxyPort, proxyPort.get)
+      System.setProperty(DisabledAuthSchemes, "")
+
+      Authenticator.setDefault(new Authenticator() {
```
patelh, line 71: Hm, is this a global?  Is there a way to isolate it for use only by this component?

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,174 @@
...
+    BigQueryOptions
+      .newBuilder()
+      .setCredentials(ServiceAccountCredentials.fromStream(new FileInputStream(gcpCredentialsFilePath)))
+      .setProjectId(gcpProjectId)
+      .setRetrySettings(ServiceOptions.getDefaultRetrySettings().toBuilder().setMaxAttempts(retries).build())
+      .build()
+      .getService
+  }
+
+  def configureProxy(): Unit = {
```
patelh, line 61: What's this proxy for?

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,174 @@
...
+
+class BigqueryQueryExecutor(
+  config: BigqueryQueryExecutorConfig,
+  lifecycleListener: ExecutionLifecycleListener
+) extends QueryExecutor
+  with Logging {
+
+  val engine: Engine = BigqueryEngine
+  config.configureProxy()
+  lazy val bigqueryClient = config.buildBigqueryClient()
```
patelh, line 89: Is this client builder meant to be lazy?  may not fail until first access but after Executor instance created successfully.

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,174 @@
...
+    if (debugEnabled) {
+      info(s"Running query : $formattedQuery")
+    }
+
+    val queryJobConfig = QueryJobConfiguration
+      .newBuilder(formattedQuery)
+      .setUseLegacySql(false)
+      .build()
+
+    val jobId = JobId.of(UUID.randomUUID.toString)
```
patelh, line 119: Could you reuse the request id here ?  Would that make sense for tracking or do you want this to be different ?

core/src/main/scala/com/yahoo/maha/core/LiteralMapper.scala
```diff
@@ -181,3 +181,28 @@ class DruidLiteralMapper extends LiteralMapper {
...
+    grain match {
+      case DailyGrain => "YYYY-MM-DD"
+      case HourlyGrain => "YYYY-MM-DD-HH"
+      case _ => "YYYY-MM-DD"
+    }
+  }
+
+  def toLiteral(column: Column, value: String, grainOption: Option[Grain] = None): String = {
+    val grain = grainOption.getOrElse(DailyGrain)
+    val escapedValue = getEscapedSqlString(value)
```
patelh, line 196: This uses the ORACLE string escaping codec for ESAPI by default.  If BigQuery has it's own implementation, you may want to use it instead.  See postgres, we use the postgres driver to do it in the postgres executor component.

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Done

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Done

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Switched from the global Authenticator to setting the proxy auth as transport options in the Google API client.

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: We can't access BigQuery from hosts in our DCs without going through a proxy server, so I included the option to configure a proxy if needed

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Had it as `lazy` initially so that we could instantiate the Executor in the ExecutoryFactory tests without needing to pass valid credentials files, but I've removed it now.

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Switched over to request ID

core/src/main/scala/com/yahoo/maha/core/LiteralMapper.scala
ishan8: Tried looking into this, but I couldn't find any specific escaping codec for BigQuery

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    builderWithCredentials
+      .build()
+      .getService
+  }
+
+  private[this] def googleApiHttpTransportFactory(): HttpTransportFactory = {
+    info(s"Setting up proxied Google API Http Transport")
+    val fileInputStream = new FileInputStream(proxyCredentialsFilePath.get)
```
patelh, line 100: This looks like an unsafe call to get(), you could do something like this:
```
private[this] def googleApiHttpTransportFactory(): Option[HttpTransportFactory] = {
  if (enableProxy && proxyCredentialsFilePath.isDefined && proxyHost.isDefined && proxyHost.isDefined) {
    info(s"Setting up proxied Google API Http Transport")
    val fileInputStream = new FileInputStream(proxyCredentialsFilePath.get)
    try {
      ...
      return Option(mHttpTransport)
    }
  } else None
```

patelh: Then you can change line 73 if block to something like this:
```
googleApiHttpTransportFactory().fold(builder.setCredentials(ServiceAccountCredentials.fromStream(fileInputStream))) {
  proxyTransport =>
    ...
}
```

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,227 @@
...
+
+  private val mapper = {
+    val om = new ObjectMapper(new YAMLFactory()) with ScalaObjectMapper
+    om.registerModule(DefaultScalaModule)
+    om.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
+    om
+  }
+
+  def buildBigqueryClient(): BigQuery = {
+    val builder = BigQueryOptions
```
patelh, line 59: You probably want to set the Connect and Read timeouts here and pass them in as config.

patelh: See HttpTransportOptions

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,227 @@
...
+      FormatConfig(dialect = SqlDialect.StandardSQL)
+    )
+
+    if (debugEnabled) {
+      info(s"Running query : $formattedQuery")
+    }
+
+    val queryJobConfig = QueryJobConfiguration
+      .newBuilder(formattedQuery)
+      .setUseLegacySql(false)
```
patelh, line 163: Could this be a config setting or is the legacy sql use deprecated ?

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
```diff
@@ -0,0 +1,227 @@
...
+          val exception = new RuntimeException("BigQuery Job Failed: " + jobId)
+          Try(lifecycleListener.failed(query, acquiredQueryAttributes, exception))
+          QueryResult(rl, acquiredQueryAttributes, QueryResultStatus.FAILURE, Option(exception))
+        } else {
+          val result = queryJob.getQueryResults()
+          val resultColumns = result.getSchema().getFields()
+          if (resultColumns.size != queryColumns.size)
+            throw new RuntimeException("Mismatch in number of query and result columns")
+
+          val rowCount = result.getTotalRows()
```
patelh, line 198: you could move this to inside the if statement below since it's only used there.

travis/publish_to_bintray.sh
```diff
@@ -18,3 +18,4 @@ mvn -q deploy -Dmaven.test.skip=true -Dscoverage.skip=true -DskipTests -Dargumen
 mvn -q deploy -Dmaven.test.skip=true -Dscoverage.skip=true -DskipTests -Darguments="-DskipTests -Dscoverage.skip=true" --projects com.yahoo.maha:maha-request-log --settings travis/settings-maha-request-log.xml
 
 mvn -q deploy -Dmaven.test.skip=true -Dscoverage.skip=true -DskipTests -Darguments="-DskipTests -Dscoverage.skip=true" --projects com.yahoo.maha:maha-postgres-executor --settings travis/settings-maha-postgres-executor.xml
+mvn -q deploy -Dmaven.test.skip=true -Dscoverage.skip=true -DskipTests -Darguments="-DskipTests -Dscoverage.skip=true" --projects com.yahoo.maha:maha-bigquery-executor --settings travis/settings-maha-bigquery-executor.xml
```
patelh, line 21: @pranavbhole probably need to add this new project to bin tray so it gets deployed correctly.

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Done

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Added these configs

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: The query generation code is using BigQuery Standard SQL, so executing as legacy SQL probably shouldn't be an option

bigquery/src/main/scala/com/yahoo/maha/executor/bigquery/BigqueryQueryExecutor.scala
ishan8: Done

patelh (approved):

patelh:
@pranavbhole probably need to add this new project to bintray so it gets deployed correctly before we merge.

pranavbhole (approved):

PR #751, opened 2021-01-11
Add logging for debugging "Querying for Future Date" for timezones ahead of utc. Fix dimQueries to not fail for the same

lgadde:

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
@@ -52,8 +52,9 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
     var utcDayFilter = localTimeDayFilter
     var utcHourFilter = localTimeHourFilter
     var utcMinuteFilter = localTimeMinuteFilter
-    if (isDebugEnabled) info(s"Timezone: $timezone")
+    info(s"Timezone: $timezone")
     if (!validateFilters(localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter) || !timezone.isDefined) {
+      info(s"validateFilters is "+ validateFilters(localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter) + "timezone is defined " + timezone.isDefined)
       if (utcHourFilter.isEmpty) {
         warn(s"Failed to validate day/hour filters, or timezone cannot be fetched. Extending day filter by one day")
         utcDayFilter = extendDaysBackwardsByOneDay(utcDayFilter)
@@ -61,6 +62,7 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
       }
     } else {
       //if timezone is UTC, pass through
+      info(s"Filters are valid and time zone is defined")
       if (timezone.contains(DateTimeZone.UTC.toString)) return (localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter)
 
       val dateTimeZone = DateTimeZone.forID(timezone.get)
@@ -286,9 +288,13 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
 
   def validateFilters(dayFilter: Filter, hourFilter: Option[Filter], minuteFilter: Option[Filter]): Boolean = {
     if (!hourFilter.isDefined && !minuteFilter.isDefined) return true
+    info(s"Hour or minute filter is not defined")
     if (!hourFilter.isDefined && minuteFilter.isDefined) return false
+    info(s"validated minuteFilter is not defined without hourFilter")
     if (hourFilter.isDefined && dayFilter.operator != hourFilter.get.operator) return false
+    info(s"hourFilter is defined both day and hour filter operator are the same")
     if (hourFilter.isDefined && minuteFilter.isDefined && hourFilter.get.operator != minuteFilter.get.operator) return false
+    info(s"minute filter operator is same as hour filter operator")
     true
   }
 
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
```diff
@@ -52,15 +52,17 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
     var utcDayFilter = localTimeDayFilter
     var utcHourFilter = localTimeHourFilter
     var utcMinuteFilter = localTimeMinuteFilter
-    if (isDebugEnabled) info(s"Timezone: $timezone")
+    info(s"Timezone: $timezone")
     if (!validateFilters(localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter) || !timezone.isDefined) {
+      info(s"validateFilters is "+ validateFilters(localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter) + "timezone is defined " + timezone.isDefined)
```
pranavbhole, line 57: store previous result of validateFilters before calling it twice? and fix string interpolations

core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
```diff
@@ -286,9 +288,13 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
 
   def validateFilters(dayFilter: Filter, hourFilter: Option[Filter], minuteFilter: Option[Filter]): Boolean = {
     if (!hourFilter.isDefined && !minuteFilter.isDefined) return true
+    info(s"Hour or minute filter is not defined")
     if (!hourFilter.isDefined && minuteFilter.isDefined) return false
+    info(s"validated minuteFilter is not defined without hourFilter")
     if (hourFilter.isDefined && dayFilter.operator != hourFilter.get.operator) return false
+    info(s"hourFilter is defined both day and hour filter operator are the same")
     if (hourFilter.isDefined && minuteFilter.isDefined && hourFilter.get.operator != minuteFilter.get.operator) return false
+    info(s"minute filter operator is same as hour filter operator")
```
ryankwagner, line 297: These logs should be clubbed.  Maybe make a secondary single log which gives you all of those values ahead of their checks.

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
```diff
@@ -355,6 +355,7 @@ object RequestModel extends Logging {
           val isDebugEnabled = request.isDebugEnabled
           //the zone should come from request or user time zone provider, not UTC time provider
           val timezone = request.getTimezone orElse userTimeZoneProvider.getTimeZone(request)
+          info(s"timezone is " +  timezone)
```
ryankwagner, line 358: The s"" is for interpolated Strings.  So timezone should be interpolated inside of that String, rather than appended.

ryankwagner: Also, as this calls both request and the provider, maybe which provider is being used can be named here rather than adding another log into the noop provider.

core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
```diff
@@ -286,9 +289,18 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
 
   def validateFilters(dayFilter: Filter, hourFilter: Option[Filter], minuteFilter: Option[Filter]): Boolean = {
     if (!hourFilter.isDefined && !minuteFilter.isDefined) return true
-    if (!hourFilter.isDefined && minuteFilter.isDefined) return false
-    if (hourFilter.isDefined && dayFilter.operator != hourFilter.get.operator) return false
-    if (hourFilter.isDefined && minuteFilter.isDefined && hourFilter.get.operator != minuteFilter.get.operator) return false
+    if (!hourFilter.isDefined && minuteFilter.isDefined) {
```
ryankwagner, line 292: all of these logs can be a single log line with condition checking inside it.

core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
```diff
@@ -52,15 +52,18 @@ class BaseUTCTimeProvider extends UTCTimeProvider with Logging {
     var utcDayFilter = localTimeDayFilter
     var utcHourFilter = localTimeHourFilter
     var utcMinuteFilter = localTimeMinuteFilter
-    if (isDebugEnabled) info(s"Timezone: $timezone")
-    if (!validateFilters(localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter) || !timezone.isDefined) {
+    info(s"Timezone: $timezone")
+    val validFilters = validateFilters(localTimeDayFilter, localTimeHourFilter, localTimeMinuteFilter)
+    if (!validFilters || !timezone.isDefined) {
+      info(s"validateFilters is  $validFilters timezone is defined ${timezone.isDefined}")
```
ryankwagner, line 58: Maybe merge line 55 down into this log, since we only care about this issue on specific requests rather than all of them.

core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
lgadde: fixed

core/src/main/scala/com/yahoo/maha/core/RequestModel.scala
lgadde: changed.

core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
lgadde: changed.

core/src/main/scala/com/yahoo/maha/core/BaseUTCTimeProvider.scala
lgadde: As discussed offline, not fixing this.

ryankwagner (approved):

PR #765, opened 2021-03-09
Add test and modification to bestCandidate Calculation

ryankwagner:
Fixes:
https://github.com/yahoo/maha/issues/766

By adding the same if-else check made into the first half of this calculation, rollup inconsistent behavior is fixed.

### DIFF
--- a/core/src/test/scala/com/yahoo/maha/core/query/QueryPipelineWithFallbackTest.scala
+++ b/core/src/test/scala/com/yahoo/maha/core/query/QueryPipelineWithFallbackTest.scala
@@ -224,42 +224,7 @@ class QueryPipelineWithFallbackTest extends AnyFunSuite with Matchers with Befor
     assert(pipeline.fallbackQueryChainOption.isEmpty, s"No fallback query expected: $pipeline")
   }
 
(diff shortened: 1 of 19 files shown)


core/src/test/scala/com/yahoo/maha/core/query/QueryPipelineWithFallbackTest.scala
```diff
@@ -223,4 +223,99 @@ class QueryPipelineWithFallbackTest extends AnyFunSuite with Matchers with Befor
     pipeline = builder._1.toOption.get.build()
     assert(pipeline.fallbackQueryChainOption.isEmpty, s"No fallback query expected: $pipeline")
   }
+
+  private[this] def aLessThanBByLevelAndCostAndCardinality(a: (String, Engine, Long, Int, Int), b: (String, Engine, Long, Int, Int)): Boolean = {
```
pranavbhole, line 227: instead of replicating code here, we can expose public wrapper method to test this comparator in the DefaultQueryPipelineFactory.

pranavbhole (approved):

PR #802, opened 2021-06-06
Add logic to return an unfiltered, versioned Domain JSON

ryankwagner:
This is allowing the Domain JSON to return multiple revisions of the same fact/dimension so that on a single query of this String, a user may view the entire contents of the registry in the same granularity as Domain.
It also lightly refactors the code so that existing functions may be reused without excess needless duplication.

A quicker revision of this code would be to put this extra variable in the same spot as these 3 are already:
`val (domainJsonAsString : String, cubesJsonStringByName: Map[String, String], cubesJson: String)`

JSON Return as such:
https://gist.github.com/ryankwagner/a38ebe14e2a2ee32194f227bb2e05a8d


core/src/main/scala/com/yahoo/maha/core/registry/Registry.scala
```diff
@@ -496,8 +506,10 @@ case class Registry private[registry](dimMap: Map[(String, Int), PublicDimension
       val maxDaysLookBack = JArray(toMaxDaysList(publicFact.maxDaysLookBack))
       val maxDaysWindow = JArray(toMaxDaysList(publicFact.maxDaysWindow))
 
+      val nameJson = if(!useRevisions) ("name" -> toJSON(publicFact.name)) else ("name,revision" -> toJSON(s"""${publicFact.name},${publicFact.revision}"""))
```
pranavbhole, line 509: we can not modify (backward incompatible change) existing domain json as it is already used by many.
you can expose new json without modifying existing one

pranavbhole (approved):

PR #812, opened 2021-06-15
Dynamic Rockdb Lookup supporting dynamic message formats and fields. : Ready to review

pranavbhole:
What is Dynamic lookup? 
Dynamic lookup supports reading the field value from rockdb record with dynamic schema,  along with rocksdb.zip, we upload the dynamic-schema.json file. Schema Manager caches the lookup => schema map and helps in the DerSer message supporting current schema.  For next snapshots, schema is updated into Manager.  For protobuf , we are creating Message descriptors on fly on reading this schema file, which is used for DeSer. For FlatBuffer, we wrote the GenericFlatBufferTable class which helps to peek into fields based on the indices defined in the schema. 

Please review!! 
Thank you 

{
  "type": "PROTOBUF",
  "version" : "2021061800",
  "name" : "product_ad_pb_dym_lookup",
  "schemaFieldList": [
    {
      "field": "id",
      "dataType": "STRING",
      "index": 1
    },
    {
      "field": "title",
      "dataType": "STRING",
      "index": 2
    },
    {
      "field": "status",
      "dataType": "STRING",
      "index": 3
    },
    {
      "field": "last_updated",
      "dataType": "STRING",
      "index": 4
    }
  ]
}


pranavbhole:
tested in the local set, works

mcjyang (approved):

PR #838, opened 2021-07-01
Add table schema parsing for calcite

ritvikjaiswal:

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
@@ -147,9 +147,9 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
         else if (fromTableAndTableSchema.size == 2)
           (fromTableAndTableSchema.get(1).toLowerCase, Some(fromTableAndTableSchema.get(0).toLowerCase))
         else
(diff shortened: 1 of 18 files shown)


service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
```diff
@@ -132,12 +134,22 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
...
+    registry.getFact(fromTable)
+  }
+
+  def getFromTableAndTableSchema(sqlNode: SqlNode): (String, Option[String]) = {
     sqlNode match {
       case sqlIdentifier: SqlIdentifier =>
-        registry.getFact(sqlIdentifier.getSimple.toLowerCase)
-      case e=>
-        throw new IllegalArgumentException(s"missing case ${e} in get cube method")
+        val fromTableAndTableSchema = sqlIdentifier.toString.toLowerCase.split("\\.")
```
pranavbhole, line 144: Instead of string parsing, SqlIndentifier provides the clean way to get the schema, i see that it is populated as part of names. You can just check on the names array length to see if schema is present

[
Screen Shot 2021-07-01 at 1 27 40 PM
](url)

pranavbhole (changes requested):

pranavbhole:
can you also add test to wrap table names with double and single quote,  try wrapping schema too

pranavbhole (approved):

PR #839, opened 2021-07-01
Fix fallback queries to use fallback engine when empty rowList is ret…

ryankwagner:
…urned.

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
@@ -125,10 +125,6 @@ case class QueryPipelineWithFallback(queryChain: QueryChain
     val stats = new EngineQueryStats
     Try {
       val result = queryChain.execute(executorContext, rowListFn, QueryAttributes.empty, stats)
-      if(result.rowList.isEmpty){
-        throw new IllegalStateException(s"Primary query returned empty rowList, attempting to fall back on ${fallbackQueryChain.drivingQuery.engine.toString}.")
-      }
-
       QueryPipelineResult(this, queryChain, result.rowList, result.queryAttributes, result.pagination)
     }.recover {
       case throwable =>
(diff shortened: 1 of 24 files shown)


core/src/main/scala/com/yahoo/maha/core/query/QueryPipeline.scala
```diff
@@ -125,6 +125,10 @@ case class QueryPipelineWithFallback(queryChain: QueryChain
     val stats = new EngineQueryStats
     Try {
       val result = queryChain.execute(executorContext, rowListFn, QueryAttributes.empty, stats)
+      if(result.rowList.isEmpty){
```
panditsurabhi, line 128: The aim is not to fallback for empty data, only for errors. There can be no rows for valid reasons and fallback is unnecessary.

druid/src/main/scala/com/yahoo/maha/executor/druid/DruidQueryExecutor.scala
```diff
@@ -413,9 +413,11 @@ class DruidQueryExecutor(config: DruidQueryExecutorConfig, lifecycleListener: Ex
       && latestDate.isBeforeNow()
       && response.getHeaders().contains(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT)
       && response.getHeader(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT).contains(DruidQueryExecutor.UNCOVERED_INTERVAL_VALUE)) {
-      //val exception = new IllegalStateException("Druid data missing, identified in uncoveredIntervals")
       logger.error(s"uncoveredIntervals Found: ${response.getHeader(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT)}")
-      //throw exception // will add-back in a week, assuming few enough intervals are found.
+      if(response.getHeader(DruidQueryExecutor.DRUID_RESPONSE_CONTEXT).contains(s""""uncoveredIntervalsOverflowed":true""")) {
```
ryankwagner, line 417: this condition in addition to a config parameter which defaults as false.

panditsurabhi (approved):

pranavbhole (approved):

PR #860, opened 2021-07-20
Maha calcite NotEqualToFilter

ritvikjaiswal:

### DIFF
--- a/service/src/test/scala/com/yahoo/maha/service/calcite/DefaultMahaCalciteSqlParserTest.scala
+++ b/service/src/test/scala/com/yahoo/maha/service/calcite/DefaultMahaCalciteSqlParserTest.scala
@@ -336,4 +336,23 @@ class DefaultMahaCalciteSqlParserTest extends BaseMahaServiceTest with Matchers 
 
     assert(request.filterExpressions.toString contains "EqualityFilter(Student ID,123,false,false)")
   }
+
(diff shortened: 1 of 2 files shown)


service/src/test/scala/com/yahoo/maha/service/calcite/DefaultMahaCalciteSqlParserTest.scala
```diff
@@ -336,4 +336,23 @@ class DefaultMahaCalciteSqlParserTest extends BaseMahaServiceTest with Matchers
 
     assert(request.filterExpressions.toString contains "EqualityFilter(Student ID,123,false,false)")
   }
+
+  test("test filter: not equals") {
+
+    val sql =
+      s"""
+              select * from student_performance
+              where 'Student ID' != 123
```
pranavbhole, line 345: can you add test for not equals with string

mcjyang (approved):

service/src/test/scala/com/yahoo/maha/service/calcite/DefaultMahaCalciteSqlParserTest.scala
ritvikjaiswal: Done

pranavbhole (approved):

PR #863, opened 2021-07-21
Less than filter

njalterio:


service/src/test/scala/com/yahoo/maha/service/calcite/DefaultMahaCalciteSqlParserTest.scala
```diff
@@ -393,4 +393,21 @@ class DefaultMahaCalciteSqlParserTest extends BaseMahaServiceTest with Matchers
     assert(request.filterExpressions.head.asValues.equals("abc"))
     assert(request.filterExpressions.toString contains "NotEqualToFilter(Student Name,abc,false,false)")
   }
+  
+    test("test filter: less than") {
+
+    val sql =
+      s"""
+              select * from student_performance
+              where 'Student Name' < '123'
```
pranavbhole, line 402: can you add test with value wrapped in double quotes

PR #864, opened 2021-07-21
Calcite parser HAVING clause

ritvikjaiswal:

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
@@ -212,23 +212,22 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
         filterList = filterList ++ constructFilters(sqlBasicCall)
       case others: AnyRef =>
         logger.error(s"sqlNode type ${sqlNode.getKind} in getSelectList is not yet supported")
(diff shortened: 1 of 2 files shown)


mcjyang (dismissed):

service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
```diff
@@ -203,16 +203,31 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
...
+    sqlHaving match {
+      case sqlBasicCall: SqlBasicCall =>
+        filterList = filterList ++ constructFilters(sqlBasicCall)
+        validate(filterList)
+      case others: AnyRef =>
+        logger.error(s"sqlNode type ${sqlNode.getKind} in getSelectList is not yet supported")
         (IndexedSeq.empty, None, None, None, 1)
+      case null =>
+        logger.debug("Empty HAVING clause.")
+        validate(filterList)
```
pranavbhole, line 230: for null having, we dont need to validate filters. We should pull validate filters outside of where and having and invoke it always.

pranavbhole (approved):

service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
ritvikjaiswal: Fixed

PR #867, opened 2021-07-21
Calcite sql filter: not like

ZhaoqiWang0605:

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
@@ -252,7 +252,10 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
       case SqlKind.BETWEEN =>
         ArrayBuffer.empty += BetweenFilter(toLiteral(operands(0)), toLiteral(operands(1)), toLiteral(operands(2))).asInstanceOf[Filter]
       case SqlKind.LIKE =>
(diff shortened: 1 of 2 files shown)


service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
```diff
@@ -252,7 +252,10 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
       case SqlKind.BETWEEN =>
         ArrayBuffer.empty += BetweenFilter(toLiteral(operands(0)), toLiteral(operands(1)), toLiteral(operands(2))).asInstanceOf[Filter]
       case SqlKind.LIKE =>
-        ArrayBuffer.empty += LikeFilter(toLiteral(operands(0)), toLiteral(operands(1))).asInstanceOf[Filter]
+        if (sqlBasicCall.getOperator.getName.contains("NOT"))
```
pranavbhole, line 255: is there any cleaner way to do this in sql cell ?
NOT can be lowercase as well.

ZhaoqiWang0605: I used to try to use the `negated` field in SqlOperator, but in SqlOperator class this field does not exist. And tried to search online as well, only find negate() method which is not fit into this case
Screen Shot 2021-07-20 at 6 21 51 PM

ZhaoqiWang0605: just found out a clearer way to do that:
`sqlBasicCall.getOperator.asInstanceOf[SqlLikeOperator].isNegated`

pranavbhole (approved):
Amazing @ZhaoqiWang0605 . 
Thank you very much

PR #868, opened 2021-07-27
[maha JDBC] get table request fix

mcjyang:
for maha, since we don't currently use catalog, return empty string for that column

Reference: 
1. https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html?is-external=true#getTables--
2. https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/adapter/jdbc/JdbcSchema.java#L263-L266

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/calcite/avatica/MahaAvaticaService.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/calcite/avatica/MahaAvaticaService.scala
@@ -18,6 +18,7 @@ import com.yahoo.maha.service.{MahaRequestContext, MahaService, RegistryConfig}
 import com.yahoo.maha.service.calcite.{DescribeSqlNode, MahaCalciteSqlParser, SelectSqlNode}
 import grizzled.slf4j.Logging
 import org.apache.commons.codec.digest.DigestUtils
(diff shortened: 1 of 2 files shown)


service/src/main/scala/com/yahoo/maha/service/calcite/avatica/MahaAvaticaService.scala
```diff
@@ -355,7 +357,15 @@ object MahaAvaticaServiceHelper extends Logging {
         }
     }
 
-    val tableMetaArray: Array[String] = Array("TABLE_NAME", "TABLE_TYPE", "REMARKS")
+    val tableMetaArray: Array[String] = Array("TABLE_CAT", "TABLE_SCHEM", "TABLE_NAME", "TABLE_TYPE", "REMARKS", "TYPE_CAT", "TYPE_SCHEM", "TYPE_NAME", "SELF_REFERENCING_COL_NAME", "REF_GENERATION")
+    val TABLE_CAT = null
```
pranavbhole, line 361: Instead of nulls, we can pass empty strings to avoid NPRs in avatica clients

pranavbhole (approved):

PR #871, opened 2021-08-02
Table Aliasing

priyankadadlani:

Adding functionality to ignore table alias in calcite parser instead of throwing an error.

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
@@ -140,7 +140,16 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
   }
 
   def getFromTableAndTableSchema(sqlNode: SqlNode): (String, Option[String]) = {
-    sqlNode match {
+    var sqlFromNode = sqlNode
+    sqlFromNode match {
+      case sqlBasicCall: SqlBasicCall =>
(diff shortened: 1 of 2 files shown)


service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
```diff
@@ -140,7 +140,16 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
...
+    var sqlFromNode = sqlNode
+    sqlFromNode match {
+      case sqlBasicCall: SqlBasicCall =>
+        sqlBasicCall.getOperator.kind match {
+          case SqlKind.AS =>
+            sqlFromNode = sqlBasicCall.getOperands() (0)
+      }
+      case _ => logger.debug("From sqlNode is not an sqlBasicCall type")
+    }
+    sqlFromNode match {
```
pranavbhole, line 152: `    val sqlFromNode  =  sqlFromNode match {
      case sqlBasicCall: SqlBasicCall =>
        sqlBasicCall.getOperator.kind match {
          case SqlKind.AS =>
            sqlFromNode = sqlBasicCall.getOperands() (0)
      }
      case _ => sqlFromNode
    }`

pranavbhole (approved):

service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
priyankadadlani: Done

PR #875, opened 2021-08-06
Setting file retention for 5 days for rocksdp cleanup

naishu:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/RocksDBManager.java
```diff
@@ -205,7 +205,7 @@ private String startNewInstance(final RocksDBExtractionNamespace extractionNames
 
         downloadRocksDBInstanceFromHDFS(hdfsPath, localZippedFileNameWithPath);
         unzipFile(localZippedFileNameWithPath);
-
+        FileRetention FileRetention = new FileRetention();
```
pranavbhole, line 208: can you make this method static?

PR #883, opened 2021-10-18
Add fallback to lookups

ritvikjaiswal:


ritvikjaiswal:
Changes included in this PR-

- Fallback to lookupService if cache is unavailable or lookup value cannot be retrieved from cache
- Emit monitoring metrics whenever this^ happens
- Emit monitoring metrics whenever cache is present but a NULL value is returned from the cache

mcjyang:
Have discussed an issue with @ritvikjaiswal, just want to write it down for visibility: 
While falling back to lookup service from a host, the targetted service hosts will reuse the extraction logic in the attempt of value fetching.  For the case of null value, it will become an infinite loop of calling lookup service within service hosts.

PR #903, opened 2022-01-06
Druid 21 fix iterable

ritvikjaiswal:


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/RocksDBLookupExtractor.java
```diff
@@ -33,4 +33,14 @@ public RocksDBLookupExtractor(RocksDBExtractionNamespace extractionNamespace, Ma
...
     }
+
+    @Override
+    public boolean canIterate() {
+        return true;
+    }
+
+    @Override
+    public Iterable<Map.Entry<String, String>> iterable() {
+        return new java.util.HashMap<String, String>().entrySet();
```
pranavbhole, line 44: can you use the static value here?

pranavbhole (approved):

PR #915, opened 2022-01-19
Add Lookup JSON String output to test, make test util

ryankwagner:

- Generate JSON parser for a dummy query using MAHA_LOOKUP.
- Create a MAHA_LOOKUP dummy class to ensure users use uniform params.
- Create a util so these functions can be reused.
- Add input column optional mapping
- Tests assert on some bad input cases

### DIFF
--- a/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
+++ b/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
@@ -126,8 +126,8 @@ public abstract class BaseRocksDBLookupExtractor<U> extends MahaLookupExtractor 
                     LOG.error("Failed to get lookup value from cache. Falling back to lookupService.");
                     serviceEmitter.emit(ServiceMetricEvent.builder().build(MonitoringConstants.MAHA_LOOKUP_GET_CACHE_VALUE_FAILURE + "_" + extractionNamespace.getNamespace(), 1));
                     String cacheByteString = new String(lookupService.lookup(new LookupService.LookupData(extractionNamespace, key, valueColumn, decodeConfigOptional)));
-                    if (dimensionOverrideMap != null && dimensionOverrideMap.containsKey(cacheByteString)) {
-                        return Strings.emptyToNull(dimensionOverrideMap.get(cacheByteString));
+                    if (secondaryColOverrideMap != null && secondaryColOverrideMap.containsKey(cacheByteString)) {
+                        return Strings.emptyToNull(secondaryColOverrideMap.get(cacheByteString));
                     }
                     return new String(lookupService.lookup(new LookupService.LookupData(extractionNamespace, key, valueColumn, decodeConfigOptional)));
                 }
@@ -148,8 +148,8 @@ public abstract class BaseRocksDBLookupExtractor<U> extends MahaLookupExtractor 
                     return null;
                 } else {
                     String cacheByteString = new String(cacheByteValue);
-                    if (dimensionOverrideMap != null && dimensionOverrideMap.containsKey(cacheByteString)) {
-                        return Strings.emptyToNull(dimensionOverrideMap.get(cacheByteString));
+                    if (secondaryColOverrideMap != null && secondaryColOverrideMap.containsKey(cacheByteString)) {
+                        return Strings.emptyToNull(secondaryColOverrideMap.get(cacheByteString));
                     }
                     return new String(cacheByteValue);
                 }
--- a/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
+++ b/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
@@ -107,10 +107,6 @@ public abstract class BaseRocksDBLookupExtractor<U> extends MahaLookupExtractor 
                 return null;
             }
 
-            if (dimensionOverrideMap != null && dimensionOverrideMap.containsKey(key)) {
-                return Strings.emptyToNull(dimensionOverrideMap.get(key));
-            }
-
             Optional<DecodeConfig> decodeConfigOptional = (decodeConfig == null) ? Optional.empty() : Optional.of(decodeConfig);
             if (!extractionNamespace.isCacheEnabled()) {
                 byte[] cacheByteValue = lookupService.lookup(new LookupService.LookupData(extractionNamespace,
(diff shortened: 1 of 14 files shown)


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
```diff
@@ -122,6 +125,10 @@ public String apply(@NotNull String key, @NotNull String valueColumn, DecodeConf
                     LOG.error("RocksDB instance is null");
                     LOG.error("Failed to get lookup value from cache. Falling back to lookupService.");
                     serviceEmitter.emit(ServiceMetricEvent.builder().build(MonitoringConstants.MAHA_LOOKUP_GET_CACHE_VALUE_FAILURE + "_" + extractionNamespace.getNamespace(), 1));
+                    String cacheByteString = new String(lookupService.lookup(new LookupService.LookupData(extractionNamespace, key, valueColumn, decodeConfigOptional)));
```
pranavbhole, line 128: We do not have to map the result from Lookup Service. Lookup service calls endpoints on peer and will already returned the mapped value. Remapping is not needed.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
```diff
@@ -140,6 +147,10 @@ public String apply(@NotNull String key, @NotNull String valueColumn, DecodeConf
                     }
                     return null;
                 } else {
+                    String cacheByteString = new String(cacheByteValue);
```
pranavbhole, line 150: can we trim the cacheByteValue before mapping?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupOperatorConversion.java
```diff
@@ -66,9 +67,15 @@ public DruidExpression toDruidExpression(PlannerContext plannerContext, RowSigna
                     final Expr columnName = inputExpressions.get(2).parse(plannerContext.getExprMacroTable()); // maha lookup name
 
                     LookupReferencesManager lrm = (LookupReferencesManager) lookupReferencesManager;
-                    String missingValue = getMissingValue(inputExpressions, plannerContext);
-
-                    if (arg.isSimpleExtraction() && lookupName.isLiteral() && columnName.isLiteral()) {
+                    String missingValue = getMissingValue(inputExpressions, plannerContext, 3, MISSING_VALUE);
+                    String extractionMap = getMissingValue(inputExpressions, plannerContext, 4, "-3");
```
pranavbhole, line 71: need to write new wrapper function to get these optional maps.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupOperatorConversion.java
```diff
@@ -66,9 +67,15 @@ public DruidExpression toDruidExpression(PlannerContext plannerContext, RowSigna
...
+                    String valueExtractionMap = getMissingValue(inputExpressions, plannerContext, 5, "-3");
+                    //TODO: Enhance by passing in KV separator & delimeter.
+                    //Also, allow passing of Map type statements: Case, KV pair, etc. ex: CASE, MAP(',','->')
+                    Map<String, String> dimensionOverrideMap = extractionMap == null || extractionMap.equals("-3") ? null : Splitter.on(",").withKeyValueSeparator("->").split(extractionMap);
```
pranavbhole, line 75: Good that we are using Guava for string to map. Makes life easy :) 
We can move separator and map separator to static constant value in this class.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupOperatorConversion.java
```diff
@@ -30,7 +31,7 @@
 
     private static final SqlFunction SQL_FUNCTION = OperatorConversions
             .operatorBuilder(DRUID_FUNC_NAME)
-            .operandTypes(SqlTypeFamily.ANY, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER)
+            .operandTypes(SqlTypeFamily.ANY, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER, SqlTypeFamily.CHARACTER)
```
pranavbhole, line 34: secondaryColOverrideMap is commonly used feature, we can move it right after missing value and push KeyOverrideMap  to last.

pranavbhole:
@ryankwagner  Thank you for your PR, this pr was deployed to Sandbox and it works. You can address the comments and we can let this in ! 💯 🥇 🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇🥇

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupOperatorConversion.java
ryankwagner: You only want it moved around inside the lookup ordering itself, but retain the same ordering on all other existing lookups (non-sql), correct?

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
ryankwagner: It looks like LookupService here only works on a simple decodeConfig which can only match a single value (not a whole Map of values).

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/BaseRocksDBLookupExtractor.java
```diff
@@ -122,6 +125,10 @@ public String apply(@NotNull String key, @NotNull String valueColumn, DecodeConf
                     LOG.error("RocksDB instance is null");
                     LOG.error("Failed to get lookup value from cache. Falling back to lookupService.");
                     serviceEmitter.emit(ServiceMetricEvent.builder().build(MonitoringConstants.MAHA_LOOKUP_GET_CACHE_VALUE_FAILURE + "_" + extractionNamespace.getNamespace(), 1));
+                    String cacheByteString = new String(lookupService.lookup(new LookupService.LookupData(extractionNamespace, key, valueColumn, decodeConfigOptional))).trim();
+                    if (secondaryColOverrideMap != null && secondaryColOverrideMap.containsKey(cacheByteString)) {
+                        return Strings.emptyToNull(secondaryColOverrideMap.get(cacheByteString));
+                    }
```
pranavbhole, line 131: In this case :  A->A_MAPPED, B->B_MAPPED.
LookupService will return the return A_MAPPED.

String cacheByteString = new String(lookupService.lookup(new LookupService.LookupData(extractionNamespace, key, valueColumn, decodeConfigOptional))).trim();
                    if (secondaryColOverrideMap != null && secondaryColOverrideMap.containsKey(cacheByteString)) {
                        return Strings.emptyToNull(secondaryColOverrideMap.get(cacheByteString));
                    }

thus we do not need above code, just return the value that you got from lookup service.

pranavbhole (approved):

PR #924, opened 2022-01-27
Add URI Namespace for File & HDFS CSV parsing

ryankwagner:

Work included is tested in local HDFS using the hdfs start-all.sh script (via hadoop docs) on HDFS 3.3.1
File used is a world cities CSV both unzipped and zipped.
Updates the RowCSVWriter to use the 4.6 version of openCSV.
Updates file with significant Druid code usage with Apache license header.

Data parsers can later be updated to include TSV, JSON, ObjectMapper, etc via:
https://www.mvndoc.com/c/io.druid.extensions/druid-lookups-cached-global/io/druid/query/lookup/namespace/class-use/UriExtractionNamespace.FlatDataParser.html

Code usage (taken from old Druid revision):
https://github.com/apache/druid/blob/master/extensions-core/lookups-cached-global/src/main/java/org/apache/druid/query/lookup/namespace/UriExtractionNamespace.java#L277


druid-lookups/pom.xml
```diff
@@ -178,6 +178,12 @@
             <version>${druid.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
pranavbhole, line 181: check if is needed

druid-lookups/pom.xml
```diff
@@ -294,6 +300,12 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.druid.extensions</groupId>
+            <artifactId>druid-hdfs-storage</artifactId>
+            <version>0.21.1</version>
```
pranavbhole, line 306: variable

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/URILookupExtractor.java
```diff
@@ -0,0 +1,52 @@
...
+import com.yahoo.maha.maha_druid_lookups.query.lookup.namespace.URIExtractionNamespace;
+import com.yahoo.maha.maha_druid_lookups.server.lookup.namespace.LookupService;
+import org.apache.druid.java.util.common.logger.Logger;
+
+import java.lang.invoke.MethodHandles;
+import java.util.List;
+import java.util.Map;
+
+public class URILookupExtractor<U extends List<String>> extends OnlineDatastoreLookupExtractor<U> {
+    private static final Logger LOG = new Logger(MethodHandles.lookup().lookupClass());
```
pranavbhole, line 32: current class name

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/URIExtractionNamespaceCacheFactory.java
```diff
@@ -0,0 +1,203 @@
...
+                                uri.toString(), id, fmt.print(lastModified), fmt.print(lastCheck));
+                        return version;
+                    }
+                } catch (NumberFormatException ex) {
+                    log.debug(ex, "Failed to get last modified timestamp. Assuming no timestamp");
+                }
+                final ByteSource source;
+                if (CompressionUtils.isGz(uriPath)) {
+                    // Simple gzip stream
+                    log.debug("Loading gz");
```
ryankwagner, line 143: Test on GZ

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/query/lookup/URILookupExtractor.java
ryankwagner: Consistent with our other extractor loggers:
```
public class JDBCLookupExtractor<U extends List<String>> extends OnlineDatastoreLookupExtractor<U> {
    private static final Logger LOG = new Logger(MethodHandles.lookup().lookupClass());
    
    public class MongoLookupExtractor<U extends List<String>> extends OnlineDatastoreLookupExtractor {
    private static final Logger LOG = new Logger(MethodHandles.lookup().lookupClass());
```

druid-lookups/pom.xml
ryankwagner: Removing from lookups, bumping version & fixing in core.

pranavbhole (approved):

PR #946, opened 2022-03-15
SQL fix for translating maha lookup expression

mcjyang:

**Summary of changes**
- fix minor issues for existing dimension extraction classes.
- add maha lookup macro to translate Druid SQL expression correctly, ex: 
```
"postAggregations": [
  {
    "type": "expression",
    "name": "p0",
    "expression": "maha_lookup(\"d0\",'druid_data_lookup','val','NA',null,'-3->UNKNOWN')",
    "ordering": null
  }
]
```
Reference:
LookupExprMacro - https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/expression/LookupExprMacro.java
LookupExprMacroTest - https://github.com/apache/druid/blob/master/server/src/test/java/org/apache/druid/query/expression/LookupExprMacroTest.java

### DIFF
--- a/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupExprMacro.java
+++ b/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupExprMacro.java
@@ -53,8 +53,10 @@ public class MahaLookupExprMacro implements ExprMacroTable.ExprMacro
         String columnStr = (String) columnExpr.getLiteralValue();
         final Expr missingValueExpr = args.get(3);
         String missingValueStr = (String) missingValueExpr.getLiteralValue();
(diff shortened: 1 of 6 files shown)


druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/MahaLookupExprMacro.java
```diff
@@ -0,0 +1,126 @@
...
+    {
+        if (args.size() < 4) {
+            throw new IAE("Function[%s] must have at least 4 arguments", name());
+        }
+
+        final Expr arg = args.get(0);
+        final Expr lookupExpr = args.get(1);
+        final Expr columnExpr = args.get(2);
+        String columnStr = (String) columnExpr.getLiteralValue();
+        final Expr missingValueExpr = args.get(3);
```
pranavbhole, line 54: missing value onwards params are optional, you can handle the null checks that we are doing in SQL OP Conversion

pranavbhole (approved):

PR #954, opened 2022-05-10
Clear non-existent uri records from cache

ZhaoqiWang0605:

### Motivation
A URI lookup is reading an input data file to update the cache. If the input data file is updated by adding new records or updating an existing records, the cache can be updated as expectation. 

However, if remove a record from the input file, then the record cannot be removed from cache, causing the record can still be queried. Thus, having this PR to resolve the issue.

### Changes Summary
In `URIExtractionNamespaceCacheFactory`, add all the records into a new map from the input file. Compare the new map with cache, removing non-existing records from the cache and update/add the rest of records into the cache.

### Tests
- [x] Unit test: check the cache should be able to remove the non-existent records
- [x] Test in Druid 21 sandbox: [link](https://docs.google.com/document/d/1R5aVMUXIC10OQK3yKbcDkT3DqDkddtp_Y1zOKy53iHo/edit?usp=sharing)

### DIFF
--- a/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/URIExtractionNamespaceCacheFactory.java
+++ b/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/URIExtractionNamespaceCacheFactory.java
@@ -27,8 +27,6 @@ import com.google.inject.Inject;
 import com.yahoo.maha.maha_druid_lookups.query.lookup.DecodeConfig;
 import com.yahoo.maha.maha_druid_lookups.query.lookup.namespace.ExtractionNamespaceCacheFactory;
 import com.yahoo.maha.maha_druid_lookups.query.lookup.namespace.URIExtractionNamespace;
-import com.yahoo.maha.maha_druid_lookups.server.lookup.namespace.cache.MahaNamespaceExtractionCacheManager;
-import com.yahoo.maha.maha_druid_lookups.server.lookup.namespace.cache.OnHeapMahaNamespaceExtractionCacheManager;
 import org.apache.druid.data.SearchableVersionedDataFinder;
 import org.apache.druid.java.util.common.IAE;
 import org.apache.druid.java.util.common.RetryUtils;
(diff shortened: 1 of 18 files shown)


ryankwagner (dismissed):
LGTM.  Please test in local against cases like Postal Code lookups & other CSVs in the Druid Example lib to find any excess edge cases.

druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/URIExtractionNamespaceCacheFactory.java
```diff
@@ -155,11 +156,22 @@ public InputStream openStream() throws IOException {
...
+                final MapPopulator.PopulateResult populator = new MapPopulator<>(
                         extractionNamespace.getNamespaceParseSpec().getParser()).populate(source,
-                        cache).getLines();
-                log.info("Finished loading %d lines for namespace [%s]", lineCount, id);
-                return version;
+                        newCache);
+
+                if (newCache.size() == populator.getEntries()) {
+                    log.info("Loaded data to a new map successfully. Cleaning previous cache and adding data in the new map to cache ...");
+                    cache.keySet().retainAll(newCache.keySet());
```
pranavbhole, line 167: can you please make sure that `cache` is instance of ConcurrentHashMap ?

ZhaoqiWang0605: sure, it is instance of `ConcurrentHashMap`. 

The `cache` in this method is coming from the following code, which call `getCacheMap()` which has abstract type `ConcurrentMap<String, U>` as its return type:
https://github.com/yahoo/maha/blob/master/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/cache/MahaNamespaceExtractionCacheManager.java#L327

And `getCacheMap()` is override in the following part which initializes with `ConcurrentHashMap`
https://github.com/yahoo/maha/blob/master/druid-lookups/src/main/java/com/yahoo/maha/maha_druid_lookups/server/lookup/namespace/cache/OnHeapMahaNamespaceExtractionCacheManager.java#L69

pranavbhole: cool

pranavbhole (approved):

PR #999, opened 2022-11-02
Convert incoming day filters to Between filter

shek1608:

* Superset passes 2 Day filters such as :
```
WHERE "Day" >= '2022-10-25 00:00:00.000000'
  AND "Day" < '2022-11-01 00:00:00.000000'
```
* So new code gathers the `from` and `to` dates from the above 2 filters, and then creates a `Between filter` for Day that maha expects, after the filters are parsed.

* Test for the above

Other changes:
* Change mapping from float to double that mahajdbc expects for DecType
* Change default from and to variable names.

### DIFF
--- a/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
+++ b/service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
@@ -13,6 +13,7 @@ import org.apache.calcite.sql.parser.{SqlParseException, SqlParser}
 import org.apache.calcite.sql.validate.{SqlConformance, SqlConformanceEnum}
 import org.apache.commons.lang.StringUtils
 import org.joda.time.{DateTime, DateTimeZone}
+import org.joda.time.format.DateTimeFormat
 
 import scala.collection.mutable.ArrayBuffer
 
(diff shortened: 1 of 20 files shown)


service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
```diff
@@ -298,7 +301,30 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
       case SqlKind.IS_NULL =>
         ArrayBuffer.empty += IsNullFilter(toLiteral(operands.get(0))).asInstanceOf[Filter]
       case SqlKind.LESS_THAN=>
-        ArrayBuffer.empty += LessThanFilter(toLiteral(operands.get(0)), toLiteral(operands.get(1))).asInstanceOf[Filter]
+        if(toLiteral(operands.get(0)).equals(DailyGrain.DAY_FILTER_FIELD)) {
```
pranavbhole, line 304: put common code behind methods

shek1608: Done

pranavbhole (approved):

PR #1000, opened 2022-11-07
Refactor date Filtering in SqlParser; add hour support; add queryRequestContext

shek1608:

* Add more datetime formats
* Add hour filters if hour column is present in group by (expect for select *)
* Add hour in cube definition for student_performance, and update/add tests


service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
```diff
@@ -188,6 +191,7 @@ case class DefaultMahaCalciteSqlParser(mahaServiceConfig: MahaServiceConfig) ext
                   sqlIdentifier.names.asScala.foreach(
                     c => {
                       val publicCol: PublicColumn = getColumnFromPublicFact(publicFact, c, columnAliasToColumnMap)
+                      mahaSqlRequestContext.queryAliasToColumnNameMap += (publicCol.alias -> publicCol.alias)
```
pranavbhole, line 194: Add native method to mahaSqlRequestContext and add entry to map. It will be better to track

service/src/main/scala/com/yahoo/maha/service/calcite/MahaSqlRequestContext.scala
```diff
@@ -0,0 +1,11 @@
+package com.yahoo.maha.service.calcite
+
+import com.yahoo.maha.core.{DailyGrain, HourlyGrain}
+import org.joda.time.{DateTime, DateTimeZone}
+
+case class MahaSqlRequestContext() {
+  var fromDate,fromHour : String = _
+  var toDate : String = DailyGrain.toFormattedString(DateTime.now(DateTimeZone.UTC))
+  var toHour : String = HourlyGrain.toFormattedString(DateTime.now(DateTimeZone.UTC))
+  val queryAliasToColumnNameMap = scala.collection.mutable.Map[String, String]()
```
pranavbhole, line 10: expose the methods to update the context vars.

pranavbhole (changes requested):
need the native methods in the context.

service/src/main/scala/com/yahoo/maha/service/calcite/MahaSqlRequestContext.scala
shek1608: Done, do review

service/src/main/scala/com/yahoo/maha/service/calcite/MahaCalciteSqlParser.scala
shek1608: Done, do review

shek1608:
> need the native methods in the context.

Added functions to update/access the variables.

pranavbhole (approved):

PR #1007, opened 2023-01-06
Add Druid capabilities for Union DataSource

ryankwagner:

Issue: Druid is only querying on a single data source.  

Without making massive changes to the code base, this allows Druid definitions anywhere in the Fact hierarchy to generate Union datasource queries.  ex:
```
{"queryType":"groupBy","dataSource":{"type":"union","dataSources":[{"type":"table","name":"fact1"},{"type":"table","name":"fact2"}]},"intervals":
...
```

Note that this change is relatively simple for Druid since Druid Native queries support this sort of logic, but SQL-based queries will need a fact-fact union generator or augmentation to serve this distinct use case.

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
@@ -299,7 +299,7 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
 
   def createDataSource(dataSource: String): DataSource = {
     if(dataSource.split(",").length <= 1)
-      new TableDataSource(dataSource)
+      new TableDataSource(dataSource.split(",").head)
     else {
       val unionSources = dataSource.split(",").toSet
       val unionSourcesObject: List[TableDataSource] = unionSources.map(source => new TableDataSource(source)).toList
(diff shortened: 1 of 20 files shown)


core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
```diff
@@ -298,11 +297,22 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
     return requestModel.startIndex < maximumMaxRows;
   }
 
+  def createDataSource(dataSource: String): DataSource = {
+    if(dataSource.split(",").length <= 1)
+      new TableDataSource(dataSource.split(",").head)
```
mcjyang, line 302: is this the same return type as `fact.underlyingTableName.getOrElse(fact.name)`?
if not, will this new return type affect any existing behavior?

ryankwagner: This converts all input from String to DataSource, which is one of the valid use cases as input to the below function (it accepts String type or dataSource type)
Here's the actual code:
https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/topn/TopNQueryBuilder.java#L158

https://github.com/apache/druid/blob/master/processing/src/main/java/org/apache/druid/query/topn/TopNQueryBuilder.java#L173

String input casts to TableDataSource, DataSource input allows use of any DataSource type.

mcjyang: Sounds good - builder accepts both String and DataSource type is the answer I'm looking for.

mcjyang (approved):

PR #1009, opened 2023-01-13
AdditionalInfo: Add User-Modifiable Derived Expression

ryankwagner:

Add a COL_W_REPLACEMENTS DerivedExpression, this will:
- By default, act the same as COL.
- When user specifies, any substring in this may be replaced in the reportingRequest within additionalParameters (if we are allowing these parameters to get parsed, API restricts this parsing heavily)

This PR demos the POC on PrestoQueryGenerator, further PRs will add Hive, Druid, Oracle, etc one at a time (for easy reverts)

### DIFF
--- a/core/src/test/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorTest.scala
+++ b/core/src/test/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorTest.scala
@@ -2,9 +2,12 @@
 // Licensed under the terms of the Apache License 2.0. Please see LICENSE file in project root for terms.
 package com.yahoo.maha.core.query.presto
 
+import com.yahoo.maha.core.CoreSchema.AdvertiserSchema
 import com.yahoo.maha.core._
 import com.yahoo.maha.core.query._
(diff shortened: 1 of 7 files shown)


core/src/test/scala/com/yahoo/maha/core/query/presto/PrestoQueryGeneratorTest.scala
```diff
@@ -1102,4 +1105,49 @@ af0.campaign_id = c1.c1_id
...
+                          ],
+                         "sortBy": [
+                           { "field": "Impressions", "order": "Desc" },
+                           { "field": "Advertiser ID", "order": "Asc"}
+                         ],
+                         "additionalParameters": {
+                            "AdditionalColumnInfo":
+                            [
+                              {"field": "keyword_id", "value": "123"},
+                              {"field": "keyword", "value": "potatoes"}
```
mcjyang, line 1133: can you remind why we need to pass two lines instead directly use:
`{"field": "keyword_id", "value": "potatoes"}`
?

mcjyang (approved):

PR #1025, opened 2023-03-17
Add Druid Lookup with user-modifiable missing value override

ryankwagner:
Currently, col additions will always return MAHA_LOOKUP_EMPTY if the inner decoded column isn't there.  
This will allow users to create their own overrides so that requests can filter on expected values like nulls or empty Strings.

### DIFF
--- a/api-example/pom.xml
+++ b/api-example/pom.xml
@@ -11,7 +11,7 @@
     <parent>
         <artifactId>maha-parent</artifactId>
         <groupId>com.yahoo.maha</groupId>
-        <version>6.130-SNAPSHOT</version>
+        <version>6.131-SNAPSHOT</version>
     </parent>
 
     <name>maha api-example</name>
--- a/api-jersey/pom.xml
+++ b/api-jersey/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <groupId>com.yahoo.maha</groupId>
         <artifactId>maha-parent</artifactId>
-        <version>6.130-SNAPSHOT</version>
+        <version>6.131-SNAPSHOT</version>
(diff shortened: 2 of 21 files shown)


mcjyang:
Just try to understand the use case here, with this new function you lookup values will either get decoded to either columnIfMatched or columnIfNotMatched, and if the chosen column is null we want to replace with some user-desired value?

Also if this is the case have you tried to hit the druid with the generated druid query? to make sure the behavior is as expected

ryankwagner:
> Just try to understand the use case here, with this new function your lookup values will either get decoded to either columnIfMatched or columnIfNotMatched, and if the chosen column is null we want to replace them with some user-desired value?
> 
> Also if this is the case have you tried to hit the druid with the generated druid query? to make sure the behavior is as expected

Yes.  The current query will return MAHA_LOOKUP_EMPTY in any returned rows, ex:
```
{
        "version": "v1",
        "timestamp": "2023-03-15T00:00:00.000Z",
        "event": {
            "School ID": "1474388",
            "Class ID": "-1",
            "Class Name": "MAHA_LOOKUP_EMPTY",
            "School Name": "Lake",
            "Impressions": 10,
            "Day": "20230315"
        }
    }
    ```
    
With the fix giving:
```
{
        "version": "v1",
        "timestamp": "2023-03-15T00:00:00.000Z",
        "event": {
            "School ID": "1474388",
            "Class ID": "-1",
            "Class Name": "",
            "School Name": "Lake",
            "Impressions": 10,
            "Day": "20230315"
        }
    }
```

mcjyang (approved):

PR #1031, opened 2023-03-30
Fix Druid ThetaSketch Columns for Size and Overrideability

ryankwagner:

we get this
    
```{"type":"filtered","aggregator":{"type":"thetaSketch","name":"woeids_unique_users","fieldName":"uniqueUserCount","size":16384,"shouldFinalize":true,"isInputThetaSketch":false},"filter":{"type":"selector","dimension":"woeids","value":"20"},"name":"woeids_unique_users"}```

we want this
based on the definition in BaseDruidQueryGeneratorTest line 124
The issue is the definition on DruidQueryGenerator line 1020-1024, which overrides the filterExpr with the queried one.
```{"type":"filtered","aggregator":{"type":"thetaSketch","name":"woeids_unique_users","fieldName":"uniqueUserCount","size":16384,"shouldFinalize":true,"isInputThetaSketch":false},"filter":{"type":"or","fields":[{"type":"selector","dimension":"woeids","value":"4563"}]},"name":"woeids_unique_users"}```

In addition, we should be able to tune thetaSketch estimator size.  This PR adds that function optionally.

### DIFF
--- a/api-example/pom.xml
+++ b/api-example/pom.xml
@@ -11,7 +11,7 @@
     <parent>
         <artifactId>maha-parent</artifactId>
         <groupId>com.yahoo.maha</groupId>
-        <version>6.133-SNAPSHOT</version>
+        <version>6.134-SNAPSHOT</version>
     </parent>
 
     <name>maha api-example</name>
--- a/api-jersey/pom.xml
+++ b/api-jersey/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <groupId>com.yahoo.maha</groupId>
         <artifactId>maha-parent</artifactId>
-        <version>6.133-SNAPSHOT</version>
+        <version>6.134-SNAPSHOT</version>
     </parent>
 
     <name>maha api-jersey</name>
--- a/bigquery/pom.xml
+++ b/bigquery/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <groupId>com.yahoo.maha</groupId>
         <artifactId>maha-parent</artifactId>
-        <version>6.133-SNAPSHOT</version>
+        <version>6.134-SNAPSHOT</version>
     </parent>
 
     <name>maha bigquery executor</name>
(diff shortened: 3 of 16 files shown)

### CODE REVIEW

File: core/src/main/scala/com/yahoo/maha/core/fact/RollupExpression.scala [hunk 1]
```diff
@@ -88,5 +89,6 @@ case class DruidHyperUniqueRollup(column: String) extends CustomRollup with With
   override val hasDerivedExpression: Boolean = true
   override lazy val sourceColumns: Set[String] = Set(column)
 }
+case class DruidThetaSketchRollup(size: Int = 16384) extends CustomRollup with WithDruidEngine
```
mcjyang, line 92: what is this default value?

core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
```diff
@@ -1002,8 +1002,8 @@ class DruidQueryGenerator(queryOptimizer: DruidQueryOptimizer
       }
     }
 
-    def getThetaSketchAggregatorFactory(outputFieldName: String, inputFieldName: String): AggregatorFactory = {
-      new SketchMergeAggregatorFactory(outputFieldName, inputFieldName, null, null, false, null)
+    def getThetaSketchAggregatorFactory(outputFieldName: String, inputFieldName: String, size: Int = 16384): AggregatorFactory = {
```
mcjyang, line 1005: same as above

pranavbhole (approved):

pranavbhole (approved):

core/src/main/scala/com/yahoo/maha/core/fact/RollupExpression.scala
ryankwagner: 16384 is what it generates by default.

core/src/main/scala/com/yahoo/maha/core/query/druid/DruidQueryGenerator.scala
ryankwagner: 16384 is what these druid queries generate when no value is specified.

mcjyang (approved):

PR #1049, opened 2023-07-20
Fix: OGB query should not render unused primitive cols at pre-outer layer

ZhaoqiWang0605:

### Motivation
The OGB query (Hive, Presto, Bigquery) cannot get an aggregated result if a derived NoopRollup metrics is queried. Because the NoopRollup column is rendered at the pre-outer layer, where the primitive columns which has rendered at the inner most layer will be rendered again at the pre-outer layer. 

### Solution
Before rendering primitive columns at the pre-outer layer, remove the primitive columns who are only used by NoopRollup columns who will be rendered at the pre-outer layer as well

### Example:
Here is a derived NoopRollup metrics:
```
HiveDerFactCol("derived_impressions", IntType(), SUM(DECODE("{device_id}", "1", "{impressions}", "0")), rollupExpression = NoopRollup)
```

Before the fix, the generated OGB query will be like the following query. Can find `CASE WHEN (device_id IN (11)) THEN 'Desktop' WHEN (device_id IN (22)) THEN 'Tablet' WHEN (device_id IN (33)) THEN 'SmartPhone' WHEN (device_id IN (-1)) THEN 'UNKNOWN' ELSE 'UNKNOWN' END device_id` was rendered at the pre-outer layer, even though only `derived_impressions` who is rendered at the same layer uses it
```
SELECT CONCAT_WS(',', CAST(NVL(mang_advertiser_name,'') AS STRING),CAST(NVL(mang_impressions,'') AS STRING),CAST(NVL(mang_derived_impressions,'') AS STRING))
FROM(
SELECT mang_advertiser_name AS mang_advertiser_name, impressions AS mang_impressions, derived_impressions AS mang_derived_impressions
FROM(
SELECT COALESCE(a1.mang_advertiser_name, 'NA') mang_advertiser_name, SUM(impressions) AS impressions, CASE WHEN (device_id IN (11)) THEN 'Desktop' WHEN (device_id IN (22)) THEN 'Tablet' WHEN (device_id IN (33)) THEN 'SmartPhone' WHEN (device_id IN (-1)) THEN 'UNKNOWN' ELSE 'UNKNOWN' END device_id, (SUM(decodeUDF(device_id, 1, impressions, 0))) AS derived_impressions
FROM(SELECT account_id, SUM(impressions) impressions, device_id
FROM s_stats_fact_underlying
WHERE (account_id = 12345) AND (stats_date >= '2023-07-18' AND stats_date <= '2023-07-25')
GROUP BY account_id, device_id

       )
ssfu0
LEFT OUTER JOIN (
SELECT name AS mang_advertiser_name, id a1_id
FROM advertiser_hive
WHERE ((load_time = '%DEFAULT_DIM_PARTITION_PREDICTATE%' ) AND (shard = 'all' )) AND (id = 12345)
)
a1
ON
ssfu0.account_id = a1.a1_id
       
GROUP BY COALESCE(a1.mang_advertiser_name, 'NA'), CASE WHEN (device_id IN (11)) THEN 'Desktop' WHEN (device_id IN (22)) THEN 'Tablet' WHEN (device_id IN (33)) THEN 'SmartPhone' WHEN (device_id IN (-1)) THEN 'UNKNOWN' ELSE 'UNKNOWN' END
) OgbQueryAlias
) queryAlias LIMIT 200
```

After the fix, the generated OGB query will be:
```
SELECT CONCAT_WS(',', CAST(NVL(mang_advertiser_name,'') AS STRING),CAST(NVL(mang_impressions,'') AS STRING),CAST(NVL(mang_derived_impressions,'') AS STRING))
FROM(
SELECT mang_advertiser_name AS mang_advertiser_name, impressions AS mang_impressions, derived_impressions AS mang_derived_impressions
FROM(
SELECT COALESCE(a1.mang_advertiser_name, 'NA') mang_advertiser_name, SUM(impressions) AS impressions, (SUM(decodeUDF(device_id, 1, impressions, 0))) AS derived_impressions
FROM(SELECT account_id, SUM(impressions) impressions, device_id
FROM s_stats_fact_underlying
WHERE (account_id = 12345) AND (stats_date >= '2023-07-18' AND stats_date <= '2023-07-25')
GROUP BY account_id, device_id

       )
ssfu0
LEFT OUTER JOIN (
SELECT name AS mang_advertiser_name, id a1_id
FROM advertiser_hive
WHERE ((load_time = '%DEFAULT_DIM_PARTITION_PREDICTATE%' ) AND (shard = 'all' )) AND (id = 12345)
)
a1
ON
ssfu0.account_id = a1.a1_id
       
GROUP BY COALESCE(a1.mang_advertiser_name, 'NA')
) OgbQueryAlias
) queryAlias LIMIT 200
```

### DIFF
--- a/core/src/main/scala/com/yahoo/maha/core/query/BigqueryHivePrestoQueryCommon.scala
+++ b/core/src/main/scala/com/yahoo/maha/core/query/BigqueryHivePrestoQueryCommon.scala
@@ -132,7 +132,7 @@ method to crawl the NoopRollup fact cols recursively and fill up the parent colu
  whose dependent source columns is/are NoopRollup column.
  All such parent noop rollup columns has to be rendered at OuterGroupBy layer
  */
-  def dfsNoopRollupCols(fact:Fact, cols: Set[(Column, String)], parentList: List[(Column, String)], noopRollupColSet: mutable.LinkedHashSet[(String, Column)]): Unit = {
+  def dfsNoopRollupCols(fact:Fact, cols: List[(Column, String)], parentList: List[(Column, String)], noopRollupColSet: mutable.LinkedHashSet[(String, Column)]): Unit = {
     cols.foreach {
       case (col, alias)=>
         col match {
@@ -166,7 +166,7 @@ method to crawl the NoopRollup fact cols recursively and fill up the parent colu
                   val sourceColAlias = sourceCol.alias.getOrElse(sourceCol.name)
                   if (col.alias.getOrElse(col.name) != sourceColAlias) {
                     // avoid adding self dependent columns
-                    dfsNoopRollupCols(fact, Set((sourceCol, sourceColAlias)), parentList++List((col, alias)), noopRollupColSet)
+                    dfsNoopRollupCols(fact, List((sourceCol, sourceColAlias)), parentList++List((col, alias)), noopRollupColSet)
                   }
               }
             }
@@ -188,7 +188,7 @@ method to crawl the NoopRollup fact cols recursively and fill up the parent colu
                   val sourceColAlias = sourceCol.alias.getOrElse(sourceCol.name)
                   if (col.alias.getOrElse(col.name) != sourceColAlias) {
                     // avoid adding self dependent columns
-                    dfsNoopRollupCols(fact, Set((sourceCol, sourceColAlias)), parentList++List((col, alias)), noopRollupColSet)
+                    dfsNoopRollupCols(fact, List((sourceCol, sourceColAlias)), parentList++List((col, alias)), noopRollupColSet)
                   }
               }
             }
@@ -210,7 +210,7 @@ method to crawl the NoopRollup fact cols recursively and fill up the parent colu
                   val sourceColAlias = sourceCol.alias.getOrElse(sourceCol.name)
                   if (col.alias.getOrElse(col.name) != sourceColAlias) {
                     // avoid adding self dependent columns
-                    dfsNoopRollupCols(fact, Set((sourceCol, sourceColAlias)), parentList++List((col, alias)), noopRollupColSet)
+                    dfsNoopRollupCols(fact, List((sourceCol, sourceColAlias)), parentList++List((col, alias)), noopRollupColSet)
                   }
               }
             }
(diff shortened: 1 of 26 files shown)


core/src/main/scala/com/yahoo/maha/core/query/BigqueryHivePrestoQueryCommon.scala
```diff
@@ -132,7 +132,7 @@ method to crawl the NoopRollup fact cols recursively and fill up the parent colu
  whose dependent source columns is/are NoopRollup column.
  All such parent noop rollup columns has to be rendered at OuterGroupBy layer
  */
-  def dfsNoopRollupCols(fact:Fact, cols: Set[(Column, String)], parentList: List[(Column, String)], noopRollupColSet: mutable.LinkedHashSet[(String, Column)]): Unit = {
```
ryankwagner, line 135: Why are we changing from Set to List?  I don't see anywhere the List change is benefiting us.

ryankwagner: I would expect we only want to use distinct for order preservation, but I don't see that functionality in use.

core/src/test/scala/com/yahoo/maha/core/query/hive/HiveQueryGeneratorV2Test.scala
```diff
@@ -1930,12 +1932,14 @@ class HiveQueryGeneratorV2Test extends BaseHiveQueryGeneratorTest {
                           "cube": "s_stats",
                           "selectFields": [
                               {"field": "Advertiser ID"},
-                              {"field": "CTR"}
```
ryankwagner, line 1933: Can this test change go into a distinct test?  We want to ensure existing tests go unharmed at worst, and improved at best.

core/src/main/scala/com/yahoo/maha/core/query/BigqueryHivePrestoQueryCommon.scala
ZhaoqiWang0605: while using Set the generated query will have random column rendering sequence, since the input in each query generator is using toSet() to convert from list to set

core/src/test/scala/com/yahoo/maha/core/query/hive/HiveQueryGeneratorV2Test.scala
ZhaoqiWang0605: pushed a new commit to separate the test

ryankwagner (approved):

PR #1050, opened 2023-09-08
Apply timezone from additionalParameter in hive query

ZhaoqiWang0605:

### Changes Summary
Add a feature for hive query which supports to apply the timezone value which defined in additionalParameter in request to specific time column in hive query. The change should be reusable for other engines.
- adding new hive derived expression `TIME_FORMAT_WITH_TIMEZONE`
- using an internal function in `TIME_FORMAT_WITH_TIMEZONE` to dynamic render is when generating query for a request
- added test cases
   - if not setting eg. `"additionalParameters": {"TimeZone": "America/New_York"}`, should use 'UTC' as default timezone
   - if set `"additionalParameters": {"TimeZone": "America/New_York"}`, should apply it in select columns part, filter part (if applicable) and group by part
   - above should be working for dimension column in both dimension table and fact table
   - if source column of the derived expression is a derived column, should be rendered correctly
- `mvn clean install` for maha is passed on local


core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala
```diff
@@ -562,6 +562,12 @@ object HiveExpression {
     def asString : String = s"from_unixtime(unix_timestamp(${s.asString}, '$fmt'), 'EEEE')"
   }
 
+  case class TIME_FORMAT_WITH_TIMEZONE(s: HiveExp, fmt: String, timezone: String = "UTC") extends BaseHiveExpression {
+    val hasRollupExpression = s.hasRollupExpression
+    val hasNumericOperation = s.hasNumericOperation
+    def asString : String = s"from_unixtime(unix_timestamp(from_utc_timestamp(from_unixtime(unix_timestamp(${s.asString}, 'yyyyMMddHH'), 'yyyy-MM-dd HH:mm:ss'), '$timezone')), '$fmt')"
```
pranavbhole, line 568: what is the use case for 3 level nested timestamp conversion? is there any option to honor another timezones?

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
```diff
@@ -1578,7 +1579,8 @@ object FilterSql {
                    columnsByNameMap: Map[String, Column],
                    engine: Engine,
                    literalMapper: SqlLiteralMapper,
-                   grainOption: Option[Grain] = None): SqlResult = {
+                   grainOption: Option[Grain] = None,
+                   request: Option[ReportingRequest] = None): SqlResult = {
```
pranavbhole, line 1583: you could directly pass ReportingRequest as it is guaranteed to be valid in filter render context

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
```diff
@@ -404,6 +414,22 @@ object QueryGeneratorHelper {
...
+  def overrideRenderedColWithTimezone(
+                           newTimeZone: Option[String]
+                           , column: DerivedColumn
+                           , query: String
+                         ): String = {
+    column match {
+      case HiveDerDimCol(_, dt, _, de, _, _, _) =>
+        de.expression match {
+          case TIME_FORMAT_WITH_TIMEZONE(_, fmt, tz) if newTimeZone.isDefined && tz != newTimeZone.get =>
+            query.replaceAll(tz, newTimeZone.get)
```
pranavbhole, line 426: can we move this at the query generation instead of using string literal replace all ?

ZhaoqiWang0605: Hi @pranavbhole , thanks for reviewing the PR :) Do you mean re-render the derived expression with the new timezone value? 
Currently, the rendering is happening while registering the table, when a request coming, we are using the rendered expression to generate the query directly according to this https://github.com/yahoo/maha/blob/master/core/src/main/scala/com/yahoo/maha/core/DerivedExpression.scala#L957. 
And for this timezone case, we would like to apply the specified timezone which defined in the additionalParameter in the request, but if make this change by re-rendering the expression, it will affect other requests who are running in parallel and want to use different timezones, since all these requests are sharing the same register facts info

core/src/main/scala/com/yahoo/maha/core/FilterOperation.scala
ZhaoqiWang0605: thanks, updated the PR and passing the reportingRequest directly

core/src/main/scala/com/yahoo/maha/core/query/QueryGenerator.scala
ZhaoqiWang0605: Hi @pranavbhole, updated the PR which instead of using the replaceAll(), it is using an internal function in TIME_FORMAT_WITH_TIMEZONE to dynamic render is when generating query for a request, pls check

ryankwagner (approved):

PR #1053, opened 2023-09-18
Apply timezone from additionalParameter in Presto query

ZhaoqiWang0605:

### Changes Summary
Add a feature for Presto query which supports to apply the timezone value which defined in additionalParameter in request to specific time column in hive query. This change is based on the changes in https://github.com/yahoo/maha/pull/1050.
(this change will make presto query has similar behavior as hive query to apply request specified timezone)
- adding new presto derived expression `TIME_FORMAT_WITH_TIMEZONE`
- using an internal function in `TIME_FORMAT_WITH_TIMEZONE` to dynamic render is when generating query for a request
- added test cases
   - if not setting eg. `"additionalParameters": {"TimeZone": "America/New_York"}`, should use 'UTC' as default timezone
   - if set `"additionalParameters": {"TimeZone": "America/New_York"}`, should apply it in select columns part, filter part (if applicable) and group by part
   - above should be working for dimension column in both dimension table and fact table
   - if source column of the derived expression is a derived column, should be rendered correctly
- `mvn clean install` for maha is passed on local


joshisanthosh (approved):

core/src/test/scala/com/yahoo/maha/core/query/hive/HiveQueryGeneratorV2Test.scala
```diff
@@ -2485,10 +2485,10 @@ class HiveQueryGeneratorV2Test extends BaseHiveQueryGeneratorTest {
...
-         |GROUP BY account_id, from_unixtime(unix_timestamp(from_utc_timestamp(from_unixtime(unix_timestamp(coalesce(utc_date, '2023-01-01'), 'yyyyMMddHH'), 'yyyy-MM-dd HH:mm:ss'), 'America/New_York')), 'yyyyMMdd')
+         |WHERE (account_id = 12345) AND (from_unixtime(unix_timestamp(from_utc_timestamp(from_unixtime(unix_timestamp(coalesce(utc_date, '20230101'), 'yyyyMMdd'), 'yyyy-MM-dd HH:mm:ss'), 'America/New_York')), 'yyyyMMdd') = '$fromDateHive') AND (from_unixtime(unix_timestamp(from_utc_timestamp(from_unixtime(unix_timestamp(utc_hour, 'yyyyMMddHH'), 'yyyy-MM-dd HH:mm:ss'), 'America/New_York')), 'yyyyMMdd') >= '$fromDateHive' AND from_unixtime(unix_timestamp(from_utc_timestamp(from_unixtime(unix_timestamp(utc_hour, 'yyyyMMddHH'), 'yyyy-MM-dd HH:mm:ss'), 'America/New_York')), 'yyyyMMdd') <= '$toDateHive')
+         |GROUP BY account_id, from_unixtime(unix_timestamp(from_utc_timestamp(from_unixtime(unix_timestamp(coalesce(utc_date, '20230101'), 'yyyyMMdd'), 'yyyy-MM-dd HH:mm:ss'), 'America/New_York')), 'yyyyMMdd')
```
pranavbhole, line 2491: can we avoid hard coding ?  this test will fail after window exceeds

pranavbhole (approved):

